Issue #29, opened 2015-08-18
ClusterExecutor.run logging blocks until PBS job completed


mwhamgenomics:
At line 48, self.proc.stdout.readline() hangs, meaning that the job ID is not logged until it's completed

mwhamgenomics:
Apparently this is something to do with qsub not flushing its output soon enough. Not really much that can be done about it in that case.

tcezard:
I think we could get that to work if we change the way executors works. We would run the qsub jobs without block and monitor the job with qstat after that.

tcezard:
This won't be an issue anymore in the new cluster

Issue #27, opened 2015-08-17
bcbio creates too many jobs 


tcezard:
bcbio jobs create 64 jobs where only 8 were expected:

``` bash
less /scratch/U008/edingen/jobs/150803_E00328_0017_BHCCV2CCXX/bcbio_jobs.pbs | grep '150803_E00328_0017_BHCCV2CCXX' | cut -d ')' -f 2  | wc -l 
64
less /scratch/U008/edingen/jobs/150803_E00328_0017_BHCCV2CCXX/bcbio_jobs.pbs | grep '150803_E00328_0017_BHCCV2CCXX' | cut -d ')' -f 2 | sort -u | wc -l 
8
```

mwhamgenomics:
There seems to have been a double-loop somewhere. Not entirely sure why, but this is not a problem in my current dev version. Will clean up the app's logging to add more debug info

Issue #26, opened 2015-08-12
Script writer modularity


mwhamgenomics:
At the moment, the bcl2fastq/fastqc/bcbio writers are specific to PBS. We should instead have a PBS base class and bcl2fastq/fastqc/bcbio helper classes that return string Bash commands. This should make it easier to port to other grid engines.

Issue #25, opened 2015-08-07
the copy of data back to the raw folder creates subfolders


tcezard:
150723_E00306_0025_BHCHK3CCXX in my raw test folder contains a dir called 150723_E00306_0025_BHCHK3CCXX which in turn contains a dir called 150723_E00306_0025_BHCHK3CCXX and so on...
I think it's the rsync of data back to the raw folder that create those subfolder

PR #30, opened 2015-08-18
Executors

### LINKED ISSUE #22: analysis-driver stuck in infinite loop if fastqc job fail
Relying on the fastqc.complete file to appear means that deamonized analysis-driver job will stick around in an infinite loop until the file appear or they're killed.
if fastqc.complete file fails to appear, there is no process to allow analysis-driver to exit gracefully.
We should check the status of the job and get a return code instead of relying on touched file.

mwhamgenomics:
Closes #22, #24, #25, #26. Closes Asana #44496088783161. Adding per-sample_id job submission for #27. Opening #29.


analysis_driver/driver.py
```diff
@@ -1,71 +1,58 @@
...
-    util.demultiplex_feedback(run_id)
+    # util.demultiplex_feedback(run_id)
 
     main_logger.info('Done')
     return 0
 
 
+def validate_run_info(run_info, sample_sheet):
+    assert run_info.barcode_len, 'No barcode found in RunInfo.xml'
+    assert sample_sheet.check_barcodes() == run_info.barcode_len, 'Barcode mismatch: %s (SampleSheet.csv) and %s (RunInfo.xml)' %\
```
tcezard: This should only raise a warning.
It is possible that the barcode in the samplesheet if smaller than the barcode in RunInfo.xml

analysis_driver/driver.py
```diff
@@ -1,71 +1,58 @@
...
+        (sample_sheet.check_barcodes(), run_info.barcode_len)
```
tcezard: We should return the barcode from the samplesheet not from the runinfo

analysis_driver/driver.py
```diff
@@ -93,86 +80,101 @@ def run_pbs(logger=None, input_run_folder=None, job_dir=None,
...
+    bcl2fastq_executor = executor.ClusterExecutor(os.path.join(job_dir, bcl2fastq_pbs_name), block=True)
+    bcl2fastq_executor.start()
+    bcl2fastq_exit_status = bcl2fastq_executor.join()
+    logger.info('Exit status: ' + str(bcl2fastq_exit_status))
+    if bcl2fastq_exit_status:
+        raise AnalysisDriverError('Bcl2fastq failed')
     
+    '''
     # Wait for fastqs to be created
     bcl2fastq_complete = os.path.join(job_dir, '.bcl2fastq_complete')
```
tcezard: This seems unnecessary now that you check that bcl2fastq is finished

analysis_driver/process_trigger.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def lock_file(dataset, status):
+    return os.path.join(cfg['input_dir'], '.' + dataset + '.' + status)
+
+
+def touch(file):
+    open(file, 'w').close()
+
+touch(lock_file('150723_esearbtaet', 'complete'))
```
tcezard: I don't think this should be here??

analysis_driver/util/__init__.py
```diff
@@ -14,14 +15,19 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
 
+class ProcessTriggerError(Exception):
+    def __init__(self, *args, **kwargs):
```
tcezard: this **init** is not necessary

Issue #19, opened 2015-07-31
Add notification system on run finished and run failed


tcezard:
We need to be able to notify <something> when certain event happen
typically the event would be analysis_driver start, finished, failed.
The notification method and recipient will change over time so we should be flexible on that part

PR #39, opened 2015-09-01
0.62 notifications

### LINKED ISSUE #36: Fastq file passed to bcbio are not the merged ones
But they're the input of the merge
I think bcbio_prepare_samples should return the merged files it generates so they can be passed to bcbio
https://github.com/EdinburghGenomics/Analysis-Driver/blob/master/analysis_driver/driver.py#L149

### LINKED ISSUE #38: Quiet mode
We should add a `--quiet` flag that turns off the stdout logger, so we can run the pipeline through cron without redirecting to a file or `/dev/null`

mwhamgenomics:
Closes #38. #19: Simplified use of notification centre. Fixes #36.


tcezard:
I would have preferred the fix for the bug we identified yesterday to be tagged independently before we add another set of change.
Unless you're convince we're not going to introduce more issue with this merge, could you tag 0.61.1 with the master adding the last bugfix you did in https://github.com/EdinburghGenomics/Analysis-Driver/commit/a6bcc3b43e3f07bd66e59dc52343f03250e3ebc5

mwhamgenomics:
Pushed https://github.com/EdinburghGenomics/Analysis-Driver/commit/6d6bd19e8fc37f865b6610575e60f98a4ec1bcb9 with the glob.glob fix, and deployed as tag 0.61.1

mwhamgenomics:
Should we take out the `#PBS -M` email notifications from the PBSWriter at this point?

tcezard:
Yes I think that a good idea

mwhamgenomics:
That should be ready for merging now.
M.

analysis_driver/notification/email_notification.py
```diff
@@ -1,56 +1,61 @@
...
-    def fail_step(self, step_name, **kwargs):
-        subject = "{} has failed"
-        body = "{} has failed"
+        msg = MIMEText(body, 'plain')
+        msg['Subject'] = subject
+        # msg['From'] = self.reporter_email
+        msg['To'] = ','.join(self.recipient_emails)
+
+        p = subprocess.Popen(
+            ['mail', '-t'],
```
tcezard: I don't understand the need to go through a subprocess to send the email.
We should use a university mail relay

PR #45, opened 2015-09-14
0.63 data output

### LINKED ISSUE #19: Add notification system on run finished and run failed

### LINKED ISSUE #31: Output of data to /PROCESSED
We should write the relevant outputs from BCBio to an appropriate directory

### LINKED ISSUE #41: Read configurations more robustly
At the moment, there is no validation of file paths. For example, if the path to an executable is wrong, `executor.proc` is set to `None`, hanging/breaking `self.proc.wait`. We should check the config file at the start of the run that all file paths are valid.
(The configuration should still be agnostic to what it contains)

### LINKED ISSUE #43: SMTP errors in EmailNotification
The connection to the email relay server doesn't seem particularly reliable, sometimes resulting in timeouts or 'connection refused' errors.

Possible solutions:
- `pass` if there is an error in EmailNotification._send_mail (not ideal, presumably)
- have the class try to reconnect a certain number of times
- try a different relay server

### LINKED ISSUE #44: Re-add ability to use intermediate directory
The RDF mount (at least at the moment) is too slow for multiple jobs. We should re-implement the ability to rsync the data over to /scratch for processing

mwhamgenomics:
Closes #19, closes #31, closes #41, fixes #43, closes #44


mwhamgenomics:
Will tag this as 0.7, as this is quite a large update

Issue #55, opened 2015-10-02
Executor overhaul


mwhamgenomics:
The usage of the Executor classes is a bit complex at the moment. An ideal situation would be:
- build up a list of Bash commands
- call, e.g, executor.execute(bash_commands, cluster=True, block=False)
- if cluster == True, then use PBSWriter/ClusterExecutor
  - if len(bash_commands) > 1, then write a  job array

We should also consolidate the Threaded exception-catching from quality_control.GenotypeValidation

tcezard:
You should allow arbitrary parameters to be passed with kwargs in the executor.
I don't you need every jobs to specify if they should be run on the cluster or locally but the executor should make that decision based on the config file.
A mechanism for bypassing this might be useful ie: force_ job_execution = 'local'

PR #57, opened 2015-10-08
0.81: Executors and int tests

mwhamgenomics:
#55: Adding ArrayExecutor and executor.execute, updating driver and GenotypeValidation accordingly. Updating Readme. Linux wildcards in dataset_scanner.scan_datasets.


analysis_driver/executor/executor.py
```diff
@@ -60,45 +93,66 @@ def run(self):
...
     def _process(self):
         """
         As the superclass, but with a qsub call to a PBS script.
         :rtype: subprocess.Popen
         """
-        if cfg['job_execution'] == 'pbs':
-            cmd = self._pbs_cmd(self.block)
-        else:
-            cmd = ['sh']
+        cmd = cfg.get('qsub', 'qsub') + ' -W block=true ' + self.cmd
```
tcezard: We're still using -W block=true which would not work with SGE ...

mwhamgenomics:
We already had a 0.81 tag - tagging this as 0.82

PR #62, opened 2015-10-23
Data output

### LINKED ISSUE #58: Output file discovery
At the moment, the pipeline iterates through sample IDs and expects to find specific files in specific places. This enforces consistency, but breaks easily if a file name configuration is wrong, and the code for it is not very clear.
1. We should use `glob.glob` to be more lenient with file names (e.g, `-ready-bam.bai` vs `-ready.bam.bai`)
2. We should discover/transfer all files that are there, then check that we have everything we expected.
3. At the moment, the driver returns a non-zero exit status if there are missing output files. Should this be the case (since the transfer of files was technically successful)?

mwhamgenomics:
Closes #58. Adds option to log to a directory. Adds support for Lims user sample ids.


analysis_driver/driver.py
```diff
@@ -67,10 +67,12 @@ def pipeline(input_run_folder):
     )
 
     valid_lanes = clarity.get_valid_lanes_from_HiseqX(run_info.flowcell_name)
+    app_logger.info('Valid lanes: ' + str(valid_lanes))
```
tcezard, line 70: That seems unnecessary. Already been logged in the function

analysis_driver/util/__init__.py
```diff
@@ -31,51 +31,38 @@ def bcbio_prepare_samples(job_dir, sample_id, fastqs, user_sample_id=None):
...
+    app_logger.info('Transferring file: ' + source)
+    if os.path.isfile(source):
+        app_logger.debug('Found file. Transferring.')
+        shutil.copyfile(source, dest)
 
-        output_file = os.path.join(output_dir, base_name)  # output_dir/sample_id.g.vcf.gz
-        app_logger.info('Looking for file: ' + source_file)
+        app_logger.debug('Generating md5 checksum')
+        md5 = hashlib.md5()
+        with open(dest, 'rb') as f:
```
tcezard, line 55: Is this going to generate md5sum for every file?

tcezard, line 55: That is likely to be very slow as it will have to read every file we transfer and this will be run on the main node

PR #66, opened 2015-11-01
Make the rsync run on the cluster

tcezard:
I think that it's all that is required but just in case can you review?

### DIFF
--- a/analysis_driver/process_trigger.py
+++ b/analysis_driver/process_trigger.py
@@ -56,6 +56,6 @@ def _transfer_to_int_dir(dataset, from_dir, to_dir, repeat_delay):
 
     # one more rsync after the RTAComplete is created. After this, everything should be synced
     sleep(repeat_delay)
-    exit_status += executor.execute([rsync_cmd], env='local').join()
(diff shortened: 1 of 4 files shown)


analysis_driver/process_trigger.py
```diff
@@ -51,11 +51,11 @@ def _transfer_to_int_dir(dataset, from_dir, to_dir, repeat_delay):
...
 
     while dataset_status(dataset) != 'transferring, rta complete':
-        exit_status += executor.execute([rsync_cmd], env='local').join()
+        exit_status += executor.execute([rsync_cmd], job_name='rsync', run_id=dataset).join()
         sleep(repeat_delay)
 
     # one more rsync after the RTAComplete is created. After this, everything should be synced
     sleep(repeat_delay)
-    exit_status += executor.execute([rsync_cmd], env='local').join()
+    exit_status += executor.execute([rsync_cmd], job_name='rsync', run_id=dataset).join()
```
mwhamgenomics: The default executor walltime is 8 hrs - should we specify something a bit higher here?

tcezard: Yes you're right we should we don't want the rsync to die because of  wall time. let's set it to 24hrs
Thanks

PR #70, opened 2015-12-07
Fix transfer dataset status

tcezard:


analysis_driver/driver.py
```diff
@@ -132,10 +132,10 @@ def demultiplexing_pipeline(dataset):
...
     exit_status += transfer_exit_status + fastqc_exit_status + md5_exit_status
 
-    if exit_status == 0:
-        ntf.start_stage('cleanup')
-        exit_status += _cleanup(run_id)
-        ntf.end_stage('cleanup', exit_status)
+    #if exit_status == 0:
+    #    ntf.start_stage('cleanup')
+    #    exit_status += _cleanup(run_id)
+    #    ntf.end_stage('cleanup', exit_status)
```
mwhamgenomics: Why are we removing the cleanup - was it breaking something?

PR #80, opened 2015-12-09
Refactor 7 12

mwhamgenomics:


analysis_driver/report_generation/report_crawlers.py
```diff
@@ -14,119 +14,134 @@
...
-            barcode_info[ELEMENT_NB_Q30_R2]=int(nb_bases_r2q30)
+        all_barcodes_per_lanes, top_unknown_barcodes_per_lanes = demultiplexing_parsers.parse_conversion_stats(conversion_xml_file)
+        nb_read_per_lane = Counter()
+        for project, library, lane, barcode, clust_count, clust_count_pf, nb_bases, nb_bases_r1_q30, nb_bases_r2_q30 in all_barcodes_per_lanes:
+            nb_read_per_lane[lane] += int(clust_count_pf)
+            barcode_info = self.barcodes_info.get('%s_%s_%s' % (self.run_id, lane, barcode))
+            barcode_info[ELEMENT_NB_READS_SEQUENCED] = int(clust_count)
+            barcode_info[ELEMENT_NB_READS_PASS_FILTER] = int(clust_count_pf)
+            # For the paired end reads
+            barcode_info[ELEMENT_NB_BASE_R1] = int(nb_bases)  # TODO: is this correct?
```
mwhamgenomics: We're setting BASES_R1 and BASES_R2 to the same thing here - is this intended?

tcezard: Yes it is, it assumes that read1 and read2 have the same read length which is always the case.

Issue #71, opened 2015-12-08
Quality control: compare genotyping 


tcezard:
most of the process is in place already but we need to fetch the genotype from the LIMS to perform the comparison

PR #96, opened 2016-02-09
Genotype validation

tcezard:
Adds genotype validation capability to AnalsyisDriver.
Should be in sync with the LIMS and the reporting App.


tcezard:
resolve #71

analysis_driver/transfer_data.py
```diff
@@ -139,7 +139,8 @@ def create_links_from_bcbio(sample_id, intput_dir, output_config, link_dir, quer
             links.append(link_file)
         else:
             app_logger.warning('No files found for pattern ' + src_pattern)
-            exit_status += 1
+            if output_record.get('required', True):
```
tcezard, line 142: This add capability to have output file referenced in the yaml file but will not yield an error or change the exit status if absent

analysis_driver/quality_control/genotype_validation.py
```diff
@@ -167,14 +176,20 @@ def _genotype_validation(self):
...
+
+        if not os.path.isfile(self.seq_vcf_file):
+            bam_file = self._bwa_alignment()
+            self._snp_calling(bam_file)
+        genotype_vcf = os.path.join(self.work_directory, self.sample_id + '_expected_genotype.vcf')
+        genotype_vcf = get_genotype_information_from_lims(self.sample_id, genotype_vcf)
+        if genotype_vcf:
+            self._rename_expected_genotype(genotype_vcf, self.sample_id)
+            self._vcf_validation(self.seq_vcf_file, genotype_vcf)
+        return self.seq_vcf_file, self.validation_results
```
mwhamgenomics: This can probably be a void method - the return values are never used

analysis_driver/config.py
```diff
@@ -143,7 +143,9 @@ def _etc_config(config_file):
 default = EnvConfiguration(
     [
         os.getenv('ANALYSISDRIVERCONFIG'),
-        os.path.expanduser('~/.analysisdriver.yaml')
+        os.path.expanduser('~/.analysisdriver.yaml'),
+        os.path.join(os.path.dirname(os.path.dirname(__file__)), 'etc', 'example_analysisdriver.yaml')
+
```
mwhamgenomics: Is this for tests? I think we had `etc` in the search path before, but removed it because the example config would never be suitable when actually running the pipeline

PR #98, opened 2016-02-09
Calculate yield from fastq

### LINKED ISSUE #97: Yield reported to the reporting app still contains the Adapter bases
The yield parsed from the conversion xml and passed on to the reporting app does not account for the bases that have been removed in the adapter trimming stage.
We need to calculate the nb bases and nb bases Q30 before and output that to file before sending the results to the reporting app.

tcezard:
Close #97 
This add new stage that calculate basic metrics from the fastq file with seqtk
A specific version of seqtk is needed to achieve this and can be found at:
https://github.com/EdinburghGenomics/seqtk
we need to add to the reporting app the following entries for run_elements:
clean_reads
clean_bases_r1
clean_bases_r2
clean_q30_bases_r1
clean_q30_bases_r2
happy to talk about the naming convention.


analysis_driver/report_generation/report_crawlers.py
```diff
@@ -112,6 +117,33 @@ def _populate_barcode_info_from_sample_sheet(self, samplesheet):
 
         self.run[ELEMENT_NUMBER_LANE] = len(self.lanes)
 
+    def _populate_barcode_info_from_seqtk_fqchk_files(self, run_dir ):
+        for run_element_id in self.barcodes_info:
+            barcode_info = self.barcodes_info.get(run_element_id)
+            self.barcodes_info[run_element_id][ELEMENT_RUN_NAME]
```
mwhamgenomics: This line seems to have no effect. Was this meant to validate the run name?

tcezard: No that does not need to be here

analysis_driver/reader/demultiplexing_parsers.py
```diff
@@ -80,3 +80,22 @@ def parse_conversion_stats(xml_file):
...
+
+
+def parse_seqtk_fqchk_file(fqchk_file, q_threshold):
+    with open(fqchk_file) as open_file:
+        first_line = open_file.readline()
+        header = open_file.readline().split()
+        all_cycles = open_file.readline().split()
+        first_cycle = open_file.readline().split()
+        nb_read = first_cycle[1]
+        nb_base = all_cycles[1]
```
mwhamgenomics: Do we only need the first two lines of data in the fqchk file? Can probably reduce the test file in that case

tests/test_reader/test_demultiplexing_parsers.py
```diff
@@ -0,0 +1,16 @@
...
+
+__author__ = 'tcezard'
+
+
+class Test_demultiplexing_stats(TestAnalysisDriver):
+
+    def test_parse_genotype_concordance(self):
+        fqchk_file = os.path.join(self.assets_path, '10015ATpool01_S1_L001_R1_001.fastq.gz.fqchk')
+        results = parse_seqtk_fqchk_file(fqchk_file, q_threshold=30)
+        print(results)
```
mwhamgenomics: The test doesn't assert anything

PR #106, opened 2016-02-23
Autodelete raw data

### LINKED ISSUE #104: Data deletion tier 1
We need to implement the first tier of data deletion. It will consist of:
- checking that a run is finished has been processed and been reviewed.
- delete the bcl files (Data directory), Logs and Thumbnail images
- archive the run folder (store it in a subfolder)
- patch the run process to change the status to deleted.

We also need to implement a check in Analysis driver to not process a deleted run.

mwhamgenomics:
Closes #104.

### DIFF
--- a/bin/delete_data.py
+++ b/bin/delete_data.py
@@ -4,6 +4,7 @@ import os
 from os.path import join
 import argparse
 import logging
+import requests
 
 sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 from analysis_driver.exceptions import AnalysisDriverError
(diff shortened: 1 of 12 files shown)


bin/delete_data.py
```diff
@@ -0,0 +1,168 @@
+__author__ = 'mwham'
+import sys
+import os
+from os.path import join
```
tcezard: can you change join to join_path for clarity

bin/delete_data.py
```diff
@@ -0,0 +1,168 @@
...
+        non_deleted_runs = query_api(
+            'runs',
+            'embedded={"run_elements":1,"analysis_driver_procs":1}',
+            'aggregate=True',
+            'max_results=' + str(self.deletion_limit)
+        )
+
+        deletable_runs = []
+        for r in non_deleted_runs:
+            review_statuses = r.get('review_statuses')
```
tcezard, line 76: Has this "review_statuses" aggregation been deployed already?
I can't seem to find it in the production server

bin/delete_data.py
```diff
@@ -0,0 +1,168 @@
...
+
+class RawDataDeleter(Deleter):
+    deletable_sub_dirs = ('Data', 'Logs', 'Thumbnail_Images')
+
+    def deletable_runs(self):
+        non_deleted_runs = query_api(
+            'runs',
+            'embedded={"run_elements":1,"analysis_driver_procs":1}',
+            'aggregate=True',
+            'max_results=' + str(self.deletion_limit)
```
tcezard: The deletion limit will act as a upper bound but not as a number of run deleted.
I don't think this is the intended behaviour.

PR #108, opened 2016-02-29
Rest communication refactor

### LINKED ISSUE #94: rest_communication refactor
Since the Rest API doesn't give any stack traces if something fails, rest_communication should log its own debug messages if something fails unexpectedly.

The module should also be moved to the top level under `analysis_driver`, and calls elsewhere in the app should go through it rather than `requests`.

mwhamgenomics:
Closes #94

### DIFF
--- a/analysis_driver/rest_communication.py
+++ b/analysis_driver/rest_communication.py
@@ -18,11 +18,14 @@ def api_url(endpoint, **query_args):
 
 def _req(method, url, **kwargs):
     r = requests.request(method, url, **kwargs)
-    req_call = ' '.join((r.request.method, r.request.path_url))
-    if r.status_code == 200:
-        app_logger.debug('%s (%s) -> %s' % (req_call, kwargs, r.content.decode('utf-8')))
+    # e.g: 'POST <url> ({"some": "args"}) -> {"some": "content"}. Status code 201. Reason: CREATED
+    report = '%s %s (%s) -> %s. Status code %s. Reason: %s' % (
+        r.request.method, r.request.path_url, kwargs, r.content.decode('utf-8'), r.status_code, r.reason
+    )
+    if r.status_code in (200, 201):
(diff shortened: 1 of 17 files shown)


analysis_driver/rest_communication.py
```diff
@@ -1,50 +1,68 @@
...
     
-def get_documents(endpoint, limit=10000, **kwargs):  # TODO: support embedding, max_results, etc.
-    url = api_url(endpoint)
-    param = []
-    url += '?max_results=%s' % limit
-    for key in kwargs:
-        param.append('"%s":"%s"' % (key, kwargs.get(key)))
-    if param:
-        url += '&where={%s}' % ','.join(param)
+def get_documents(endpoint, limit=10000, **query_args):
```
tcezard: If we're only allowing limited number of kwargs (where, embedded, and aggregate), shouldn't we specify them in get_document.
That would also help in knowing what paramter should be used.

analysis_driver/rest_communication.py
```diff
@@ -1,50 +1,68 @@
...
+        r.request.method, r.request.path_url, kwargs, r.content.decode('utf-8'), r.status_code, r.reason
+    )
+    if r.status_code in (200, 201):
+        app_logger.debug(report)
+    else:
+        app_logger.error(report)
     return r
 
+
+def depaginate_documents(endpoint, **queries):
```
tcezard: I'm wondering if depaginate_documents and get_documents should be merged. 
You can have get_documents with specified page number or all the pages is page=-1That would also help reduce the default value for limit

mwhamgenomics: Looks like they can be merged. You already can specify `page=` in the keyword args and it will be passed to Eve (we just haven't been using it). A boolean `depaginate` argument might be cleaner than `page=-1`

PR #120, opened 2016-03-07
Contamination check ke

katieemelianova:


bin/run_species_contamination_check.py
```diff
@@ -0,0 +1,41 @@
+import os
```
tcezard: This file Should probably go ?

analysis_driver/reader/demultiplexing_parsers.py
```diff
@@ -93,9 +103,78 @@ def parse_seqtk_fqchk_file(fqchk_file, q_threshold):
...
+def parse_fastqscreen_file(filename, myFocalSpecies):
+    """
+    parse the fastq screen outfile
+    :return dict: the maximum number of reads mapped uniquely (singly or multiple times) to a contaminant species
+    :return float: % reads unmapped to focal Species
+    :return float: % reads with no hits to any of the genomes provided
+    :return int: number of reads mapped in total
+    """
+    file = open(filename)
+    lines = file.readlines()
```
tcezard: We should use the construct:

``` python
with open(filename) as open_file:
    #Do something with the file
```

And possibly:

``` python
for line in open_file:
    #Do something with the line
```

analysis_driver/reader/demultiplexing_parsers.py
```diff
@@ -93,9 +103,78 @@ def parse_seqtk_fqchk_file(fqchk_file, q_threshold):
...
+            #header are %Q2
             if int(h[2:]) < q_threshold:
                 lo_q += int(all_cycles[9+i])
             else:
                 hi_q += int(all_cycles[9+i])
-        return nb_read, nb_base, lo_q, hi_q
+        return  nb_read, nb_base, lo_q, hi_q
+
+
+def get_focal_species(sample_id):
```
tcezard: Is this function really necessary?

analysis_driver/reader/demultiplexing_parsers.py
```diff
@@ -93,9 +103,78 @@ def parse_seqtk_fqchk_file(fqchk_file, q_threshold):
...
+            speciesName = speciesName.replace('_',' ')
+            speciesResults = result.split('\t')[1:12]
+
+            if speciesName != myFocalSpecies:
+                numberUniquelyMapped = int(result.split('\t')[4]) + int(result.split('\t')[6])
+                contaminantsUniquelyMapped[speciesName] = numberUniquelyMapped
+            elif speciesName == myFocalSpecies:
+                focalSpeciesPercentUnmapped = float(speciesResults[2])
+        contaminantsUniquelyMapped = {k:v for k,v in contaminantsUniquelyMapped.items() if v != 0}
+        ELEMENT_SPECIES_CONTAMINATION = {ELEMENT_CONTAMINANT_UNIQUE_MAP:contaminantsUniquelyMapped,
```
tcezard: I don't think this is how `ELEMENT_SPECIES_CONTAMINATION` was meant to be used.
Uppercase variables are meant to be constant and here this one is reassigned with a dict

bin/run_qc.py
```diff
@@ -32,6 +34,12 @@ def _parse_args():
     geno_val_parser.add_argument('--sample_id', required=True)
     geno_val_parser.set_defaults(func=run_genotype_validation)
 
+    sp_contamination_parser = subparsers.add_parser('contamination_check')
+    sp_contamination_parser.add_argument('--fastq_files', required=True, nargs='+', help='the fastq file pairs')
+    sp_contamination_parser.add_argument('--run_id', required=True)
```
tcezard: `--run_id` should be replace by `--working_dir` 
or better: make --working_dir optional and build it from `os.path.join(cfg[jobs_dir], sample_id`

bin/run_qc.py
```diff
@@ -82,5 +90,19 @@ def run_genotype_validation(args):
         raise AnalysisDriverError("Copy of the results files to remote has failed")
 
 
+def run_species_contamiantion_check(args):
+    work_dir = os.path.join(cfg['jobs_dir'], args.run_id)
+    os.makedirs(work_dir,exist_ok=True)
+    species_contamination_check = ContaminationCheck(sorted(args.fastq_files),args.run_id)
```
tcezard: use `working_dir` instead of `run_id`

analysis_driver/quality_control/contamination_checks.py
```diff
@@ -0,0 +1,89 @@
...
+        """
+        :return list: the command used to run fastq screen
+        """
+        if len(self.fastq_files) == 1:
+
+            fastqscreen_bin = (self.tools.get('fastqscreen'))
+            confPath = (self.contamination_cfg.get('fastqscreen_conf'))
+            fastq1 = self.fastq_files[0]
+
+            fastqscreen_command = ["{} --aligner bowtie2 {} --conf {} --force".format(fastqscreen_bin,
```
tcezard: why do this need to return an array?
isn't a string enough?

tests/test_quality_control/test_contamination_checks.py
```diff
@@ -0,0 +1,51 @@
...
+        assert my_fastq_screen_command == ["path/to/fastqscreen " \
+                                          "--aligner bowtie2 fastqFile1.fastq fastqFile2.fastq " \
+                                          "--conf path/to/fastqscreen/conf --force"]
+
+    def test_get_expected_outfiles(self):
+        fastq_files = ['fastqFile1.fastq', 'fastqFile2.fastq']
+        working_dir = 'testSample'
+        c = ContaminationCheck(fastq_files=fastq_files, working_dir=working_dir)
+        fastqscreen_expected_outfiles = c._get_expected_outfiles()
+        assert fastqscreen_expected_outfiles == ['fastqFile1_screen.txt', 'fastqFile2_screen.txt']
```
tcezard: What about when only one fastq file is provided?

tests/test_quality_control/test_contamination_checks.py
```diff
@@ -0,0 +1,51 @@
...
+    def test_run_fastqscreen(self, mocked_execute):
+        fastq_files = ['fastqFile1.fastq', 'fastqFile2.fastq']
+        working_dir = 'testSample'
+        c = ContaminationCheck(fastq_files=fastq_files, working_dir=working_dir)
+        instance = mocked_execute.return_value
+        instance.join.return_value = 0
+        run_fastqscreen = c._run_fastqscreen()
+        assert run_fastqscreen == ['fastqFile1_screen.txt', 'fastqFile2_screen.txt']
+
+    @patch('analysis_driver.executor.execute', autospec=True)
```
tcezard: That does not need to be a different test function.
It can just be a second assert in the previous test

Issue #88, opened 2016-01-19
Filter very low quality reads and report the results consistently with Conversion.xml file


tcezard:
Some of the reads we sequence end up being so low in quality that they a no chance of being useful at all. We should remove them from the raw dataset before they're passed on to the Sample processing pipeline.
We should filter the reads that have mean quality lower than 10
Count how many reads, bases, and base quality are filtered 
The resulting fastq files will be used as raw data and we need to record in a file how much data is left in each fastq files

tcezard:
This issue has been made more pressing because the adapter dimer are not removed by bcl2fastq but replaced by place holders.

tcezard:
This is now done

PR #124, opened 2016-03-12
Add quality filtering

tcezard:
This PR adds a filtering steps after demultiplexing to remove partial adapters before the metrics are extracted with seqtk fqchk


tcezard:
close #88

PR #132, opened 2016-03-22
Data deletion tier 2

### LINKED ISSUE #117: Deletion of data fails when one of the deleted folder is missing
When a run is failed or aborted some directories are missing and the deletion fails.
We need to make the move of these directories conditional of their existence.

### LINKED ISSUE #121: Data deletion tier 2
We need to implement the second tier of data deletion.
This involves:
- Check that a sample has been marked as useable and delivered (REST API)
- Check that the sample has been delivered on the LIMS
- Retrieve the run_elements associated with this sample
- Find that fastq files and delete them

mwhamgenomics:
Closes #117, closes #121


bin/delete_data.py
```diff
@@ -123,37 +132,179 @@ def run_deletion(self):
...
+                where=where,  # TODO: do we want useable only?
+                projection={'sample_id': 1},
+                depaginate=True
+            )
+        return [s['sample_id'] for s in self._samples_released_in_app]
+
+    def find_fastqs_for_run_element(self, run_element):
+        return util.find_fastqs(
+            p_join(self.data_dir, run_element['run_id'], 'fastq'),
+            run_element['project_id'],
```
tcezard: Should we use the constant here?

PR #140, opened 2016-04-11
Median coverage

### LINKED ISSUE #110: Coverage/callable bases calculation
BCBio reports a median coverage in one of its file but it is in some case very inaccurate.
We should run the coverage calculation ourselves after bcbio is done and also add it to the QC pipeline.

katieemelianova:
I havent finished writing tests e.g. for demultiplexng parsers and havent added code to report crawlers yet! closes #110


analysis_driver/constants.py
```diff
@@ -36,6 +36,7 @@
 ELEMENT_MEDIAN_INSERT_SIZE = 'Median Insert Size'
 ELEMENT_MEAN_INSERT_SIZE = 'Mean Insert Size'
 ELEMENT_MEDIAN_COVERAGE = 'median_coverage'
+ELEMENT_GATK_MEDIAN_COVERAGE = 'gatk_median_coverage'
```
tcezard: maybe not call it gatk since it's calculated with samtools?

analysis_driver/quality_control/median_coverage.py
```diff
@@ -0,0 +1,74 @@
...
+from analysis_driver.notification import default as ntf
+from .quality_control_base import QualityControl
+
+
+class SamtoolsDepth(QualityControl):
+    def __init__(self, dataset, working_dir, bam_file):
+        super().__init__(dataset, working_dir)
+        self.bam_file = bam_file
+        self.working_dir =  working_dir
+        self.exception = None
```
tcezard: exception is already defined in QualityControl

analysis_driver/quality_control/median_coverage.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+    def _get_samtools_depth_command(self):
+        samtools_bin = cfg['tools']['samtools']
+        samtools_depth_out_file = ((self.bam_file).rstrip('bam') + 'depth')
```
tcezard: I don't think these parenthesis are neccessary

analysis_driver/quality_control/median_coverage.py
```diff
@@ -0,0 +1,74 @@
...
+        """
+        :return string: the expected outfile from samtools depth
+        """
+        samtools_depth_command, samtools_depth_out_file = self._get_samtools_depth_command()
+        ntf.start_stage('run_samtools_depth')
+        samtools_depth_executor = executor.execute(
+            [samtools_depth_command],
+            job_name='samtoolsdepth',
+            working_dir=self.working_dir,
+            cpus=2,
```
tcezard: samtools is single-threaded bring dow the number of cpu to 1 and mem probably to 6

analysis_driver/quality_control/median_coverage.py
```diff
@@ -0,0 +1,74 @@
...
+        )
+        exit_status = samtools_depth_executor.join()
+        ntf.end_stage('run_samtools_depth', exit_status)
+        return samtools_depth_out_file
+
+    def _get_depth_histogram_command(self):
+        samtools_depth_outfile = self._run_samtools_depth()
+        depth_histogram_outfile = samtools_depth_outfile.rstrip('depth') + 'hist'
+        depth_histogram_command = "awk -F '\t' '{print $3}' %s | sort | uniq -c | sort -nr | sort -k2 -n > %s" % (samtools_depth_outfile,
+                                                                                                    depth_histogram_outfile)
```
tcezard: What about:

``` bash
awk -F '\t' '{array[$1"\t"$3]+=1} END{for (val in array){print val"\t"array[val]}}' | sort -k 1,1 -nk 2,2
```

The format would be a bit different but it would be faster and you would have all the chromosomes  
Alternatively 

``` bash
awk -F '\t' '{array[$3]+=1} END{for (val in array){print array[val]"\t"val}}' | sort -nk 2,2
```

would give the same results but faster

analysis_driver/quality_control/median_coverage.py
```diff
@@ -0,0 +1,74 @@
...
+        return depth_histogram_command, depth_histogram_outfile
+
+    def _run_depth_histogram_command(self):
+        depth_histogram_command, depth_histogram_outfile = self._get_depth_histogram_command()
+        ntf.start_stage('run_depth_histogram')
+        depth_histogram_executor = executor.execute(
+            [depth_histogram_command],
+            job_name='depthhistogram',
+            working_dir=self.working_dir,
+            cpus=2,
```
tcezard: no need for 2 cpus

analysis_driver/reader/demultiplexing_parsers.py
```diff
@@ -214,3 +215,77 @@ def get_fastqscreen_results(filename, sample_id):
...
+    numberOfDepths = 0
+    with open(histogram) as openfile:
+        lines = openfile.readlines()
+        for line in lines:
+            count = int(line.split()[2])
+            depth = int(line.split()[1])
+            sumOfDepths += int(count * depth)
+            numberOfDepths += int(count)
+    meanDepth = sumOfDepths/numberOfDepths
+    return math.floor(meanDepth)
```
tcezard: why rounding the mean and std dev?

tests/assets/test_crawlers/ConversionStats2.xml
```diff
@@ -0,0 +1,1060 @@
+<?xml version="1.0" encoding="utf-8"?>
```
tcezard: Why is this being added?

PR #144, opened 2016-04-17
Expand Raw data deletion to any run specified on the command line.

tcezard:
These will come on top of the run to delete found

### DIFF
--- a/bin/delete_data.py
+++ b/bin/delete_data.py
@@ -54,27 +54,18 @@ class RawDataDeleter(Deleter):
         self.list_runs = list_runs
 
     def deletable_runs(self):
-        runs = rest_communication.get_documents(
-            'runs',
-            depaginate=True,
-            max_results=100,
-            embedded={ELEMENT_RUN_ELEMENTS: 1, ELEMENT_PROCS: 1},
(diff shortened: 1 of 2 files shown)


bin/delete_data.py
```diff
@@ -64,13 +65,16 @@ def deletable_runs(self):
...
+            if self.list_runs and r[ELEMENT_RUN_NAME] in self.list_runs:
+                deletable_runs.append(r)
+            else:
+                review_statuses = r.get('review_statuses')
+                most_recent_proc = r.get(ELEMENT_PROCS, [{}])[-1]
+                if type(review_statuses) is list:
+                    review_statuses = [s for s in review_statuses if s]
+                if review_statuses and 'not reviewed' not in review_statuses:
+                    if most_recent_proc.get(ELEMENT_STATUS) in ('finished', 'aborted'):  # i.e. not 'deleted'
+                        deletable_runs.append(r)
```
mwhamgenomics: What do we want the behaviour of `--list_runs` to be? Should it force-delete runs, or select specific runs to delete?

At the moment, if you tell it to delete a specific run, it will pick up that, but also all other deletable runs anyway

PR #147, opened 2016-04-21
Batch dataset scanning

### LINKED ISSUE #102: Perform batch call to REST API for sample/run report
There is no need to run each sample/run api call independently, both the reporting-app and the LIMS API support batch call.
We'll need to refactor the Datasets and scanner a bit to get this to work

### LINKED ISSUE #116: Any unignored directory in the input dir gets processed
E.g, if someone creates an empty directory in the input location, it will be treated as a BCL dataset. We should check whether the correct subdirectories are present (Data, InterOp, Logs, etc.) before the `Dataset` object is created.

mwhamgenomics:
Closes #102, closes #116.

dataset_scanner has been refactored to allow Datasets to be created with pre-computed data from batch API calls. The initial dataset scan is now quicker, which should help keep the pipeline responsive with large numbers of runs and samples.


analysis_driver/clarity/__init__.py
```diff
@@ -96,41 +98,63 @@ def get_species_information_from_ncbi(species):
...
+    lims = connection()
+    samples = lims.get_samples(name=_sample_names)
+    lims.get_batch(samples)
+
+    if len(samples) != len(sample_names):  # haven't got all the samples because some had _01/__L:01
+        if sub < len(substitutions):
+            remainder = sorted(set(_sample_names).difference(set([s.name for s in samples])))
+            samples.extend(get_list_of_samples(remainder, sub + 1))
+        else:
+            raise LimsCommunicationError('Expected %s back, got %s' % (_sample_names, len(samples)))
```
tcezard: No That I see this, I realise that this is going to fail (and crash) in case we have samples with the same names.
Is this what we want ?

PR #150, opened 2016-04-28
Data deletion tier 3

### LINKED ISSUE #146: Data deletion tier 3
We need to:
- check the Rest API for samples that have been reviewed and delivered
- check the Lims for samples where the most recent EG Data Release 1.0 is more than 3 months ago
- delete delivered_projects/project_id/release_date/sample_id

There will eventually be a user-notification step in the Lims. Once this is in, check that the sample has been through this.

mwhamgenomics:
Closes #146


bin/delete_data.py
```diff
@@ -257,11 +253,80 @@ def delete_data(self):
...
+                {'status': 'deleted'},
+                'proc_id', proc_id
+            )
+
+    def setup_samples_for_deletion(self, sample_records):
+        for s in sample_records:
+            self._setup_sample_for_deletion(s['project_id'], s['sample_id'])
+
+    def _setup_sample_for_deletion(self, project_id, sample_id):
+        release_folders = util.find_files(cfg['data_deletion']['delivered_data'], project_id, '*', sample_id)
```
tcezard: isn't this suppose to be self.data_dir instead of cfg['data_deletion']['delivered_data']

mwhamgenomics:
I've just realised that this should probably be merged into `New_cluster_support` now, rather than `master`. Could close this req and reopen?
Edit: we're merging `New_cluster_support` into master, so just continue as before

tcezard:
This has moved to EGCG-Project-Management

PR #152, opened 2016-05-06
Add lane duplicate calculation to the run stage

tcezard:


analysis_driver/constants.py
```diff
@@ -24,6 +24,7 @@
 ELEMENT_NB_BASE_R2_CLEANED = 'clean_bases_r2'
 ELEMENT_NB_Q30_R1_CLEANED = 'clean_q30_bases_r1'
```
katieemelianova, line 25: Needs the latest reporting app pull request to be put into production before duplicate calculation can be used

tcezard:
There is a problem in testing this PR because well duplicate search for all the tiles and will crash if one is missing.
We need to add -t 1101 to the command line to make it work with test runs.

tcezard:
This PR was tested on Ultra and works.
Can this be reviewed?

katieemelianova:
Hi Tim,
Sorry it slipped my mind yesterday, I'll do that today!
Best,
Katie

Date: Mon, 9 May 2016 22:12:11 -0700

PR #155, opened 2016-05-16
Asana notification

### LINKED ISSUE #138: AsanaNotification
We should add a Notification class which if something goes wrong:
- [ ] creates an Asana ticket in the right project if it doesn't exist, otherwise retrieves it
- [ ] comments on it

This should allow us to assign failed runs to each person and comment on them, rather than reply to email notifications. There's an [API interface](https://github.com/Asana/python-asana) developed by the Asana team we should be able to use.

mwhamgenomics:
Closes #138.


analysis_driver/notification/asana_notification.py
```diff
@@ -0,0 +1,47 @@
...
+        self._add_comment('Crash report:\n\n' + crash_report)
+
+    def _add_comment(self, text):
+        self.client.tasks.add_comment(self.task['id'], text=text)
+
+    def _get_entity(self, collection, name):
+        for e in collection:
+            if e['name'] == name:
+                return e
+        self.warning('Task not found: %s' % name)
```
tcezard: That should probably not be a warning since every task creation is going to go through it.

tcezard:
This looks good, I think it's ready to merge

PR #157, opened 2016-05-17
NCBI cache

### LINKED ISSUE #148: Store species aliases localy
Recently efetch and esearch from NCBI have been often unavailable.
We should setup a local repository of species aliases that we would populate with any new alias we encounter.

mwhamgenomics:
Closes #148.

This refactors the external data source modules (`clarity`, `rest_communication` and `ncbi`) into a package, and adds a sqlite database in front of the NCBI Eutils query. If species information already exists in the cache, then no query is needed.


tcezard:
This looks good

PR #159, opened 2016-05-26
Slurm script writer

mwhamgenomics:
Adds compatibility with Slurm for #128. Also refactors the arguments to `executor.execute`:

``` python
execute(cmd)
execute(cmd1, cmd2, cmd3)
```

instead of

``` python
execute([cmd])
execute([cmd1, cmd2, cmd3])
```

PR #163, opened 2016-05-31
Slurm script writer

tcezard:
copy of #159


tcezard:
This has been reviewed already

PR #173, opened 2016-07-04
Non-human variant calling

mwhamgenomics:
This adds a GATK variant calling pipeline that is triggered by a Lims UDF. Also adds read group information to Bam files produced by `bwa mem`.

### DIFF
--- a/analysis_driver/driver.py
+++ b/analysis_driver/driver.py
@@ -21,13 +21,14 @@ def pipeline(dataset):
     if isinstance(dataset, RunDataset):
         return demultiplexing_pipeline(dataset)
     elif isinstance(dataset, SampleDataset):
-        species = clarity.get_species_from_sample(dataset.name)
+        species = clarity.get_species_from_sample(dataset.name).replace(' ', '_')
         if species is None:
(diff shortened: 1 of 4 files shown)


analysis_driver/driver.py
```diff
@@ -366,6 +385,145 @@ def qc_pipeline(dataset, species):
...
+        mem=16
+    ).join()
+    tabix = executor.execute(
+        '%s -p vcf %s' % (cfg['tools']['tabix'], sample_gvcfgz),
+        job_name='tabix',
+        working_dir=gatk_run_dir,
+        cpus=16,
+        mem=16
+    ).join()
+
```
mwhamgenomics, line 502: These cpu/mem values can probably be increased if we want.

analysis_driver/driver.py
```diff
@@ -366,6 +383,145 @@ def qc_pipeline(dataset, species):
...
```
tcezard, line 502: bgzip and tabix are single threaded so no need to give 16 cpus 
mem should alsp be reduce to 4G

PR #178, opened 2016-07-19
Het hom ti tv

### LINKED ISSUE #129: Calculate Het to hom ratio
Het to hom ratio can be used to estimate contamination.
A ratio >2 can indicate a potential contamination by a different (same species sample)
We can calculate the het/hom ratio from the vcf file.

tcezard:
This PR calculate stats from the filtered vcf file that bcbio calculates and upload ti_tv_ratio to the rest api
closes #129


analysis_driver/driver.py
```diff
@@ -254,6 +254,10 @@ def bcbio_var_calling_pipeline(dataset):
     gender_validation = qc.GenderValidation(dataset, sample_dir, vcf_file)
     gender_validation.start()
 
+    #vcf stats
+    vcf_stats = qc.VCFStats(dataset, sample_dir, vcf_file)
```
mwhamgenomics, line 261: `VCFStats` is not imported in `quality_control.__init__.py`, so this will break.
EDIT: this is now fixed.

PR #179, opened 2016-07-19
Report tools versions

### LINKED ISSUE #177: record and write out all software versions
We need to create a parsable record of all software version that were used during the analysis and make it part of the final output

tcezard:
Add version of all the software that were used to process the samples
closes #177


analysis_driver/reader/version_reader.py
```diff
@@ -0,0 +1,73 @@
...
+import yaml
+
+from analysis_driver.exceptions import AnalysisDriverError
+from analysis_driver.config import default as cfg
+
+
+class Version_reader():
+    tool_name = None
+    command = None
+    def __init__(self, tool_name, command, prefix=None):
```
mwhamgenomics: `prefix` is never used to set `self.prefix`, so `_get_stdout_from_command` will crash.

tests/test_reader/test_version_reader.py
```diff
@@ -0,0 +1,21 @@
+from unittest.mock import patch
+
+from analysis_driver.reader.version_reader import Version_reader, get_versions
+from analysis_driver.config import default as cfg
+from tests.test_analysisdriver import TestAnalysisDriver
+
+
+class TestVersion(TestAnalysisDriver):
+
```
mwhamgenomics: A test for `_get_stdout_from_command` wouldn't go amiss here.

tests/test_reader/test_version_reader.py
```diff
@@ -0,0 +1,31 @@
...
+        with patch('analysis_driver.reader.version_reader.Version_reader._get_stdout_from_command', return_value='1.2'):
+            reader = Version_reader(tool_name='bwa', command=' -h | grep "Version')
+            assert reader.get_version() == '1.2'
+
+        reader = Version_reader(tool_name='bwa', command=None)
+        assert reader.get_version() == '0.9'
+
+    def test_get_stdout_from_command(self):
+        reader = Version_reader(tool_name='echo', command=' 123')
+        reader._get_stdout_from_command() == '123'
```
mwhamgenomics: Maybe assert this instead of just declaring it!

PR #184, opened 2016-09-16
Report bases at coverage

katieemelianova:
Adding in percentiles at 25, 50, 75 and 95, and calculating number of bases covered at at least 5, 15 and 30 X.


tcezard (approved):
This looks good

PR #187, opened 2016-10-10
Transfer lims info

tcezard:
Add information taken from the LIMS to the sample in the REST API
It adds it both at the Run and at the Sample processing stage to allow to update in case anything has changed on the LIMS


analysis_driver/report_generation/report_crawlers.py
```diff
@@ -22,7 +22,27 @@
     ELEMENT_GENOTYPE_VALIDATION, ELEMENT_COVERAGE_STATISTICS, ELEMENT_MEAN_COVERAGE, ELEMENT_COVERAGE_PERCENTILES, \
     ELEMENT_BASES_AT_COVERAGE, ELEMENT_MEDIAN_COVERAGE_SAMTOOLS, ELEMENT_COVERAGE_SD, ELEMENT_FREEMIX, ELEMENT_SAMPLE_CONTAMINATION, \
     ELEMENT_GENDER_VALIDATION, ELEMENT_GENDER_HETX, ELEMENT_LANE_PC_OPT_DUP, ELEMENT_GENDER_COVY, ELEMENT_SNPS_TI_TV, \
-    ELEMENT_SNPS_HET_HOM
+    ELEMENT_SNPS_HET_HOM, ELEMENT_SAMPLE_PLATE, ELEMENT_SAMPLE_SPECIES, ELEMENT_SAMPLE_EXPECTED_YIELD, ELEMENT_SAMPLE_EXPECTED_COVERAGE
+
+
+
+def get_sample_information_from_lims(sample_name):
+    lims_sample = clarity.get_sample()
```
katieemelianova: Is this supposed to use 'sample_name' as an argument?

analysis_driver/report_generation/report_crawlers.py
```diff
@@ -22,7 +22,27 @@
...
+    gender = SampleCrawler.gender_aliases(clarity.get_sample_gender())
```
mwhamgenomics: This will raise `TypeError: 'dict' object is not callable`, and no sample name is passed to `get_sample_gender`. It should probably be something like:
`gender = SampleCrawler.gender_aliases[clarity.get_sample_gender(sample_name)]`

It would also make sense to move the gender alias out of `SampleCrawler` back to the top level.

analysis_driver/report_generation/report_crawlers.py
```diff
@@ -22,7 +22,27 @@
...
+    plate_id, well = clarity.get_plate_id_and_well(sample_name)
+    species = clarity.get_species_from_sample(sample_name)
+    external_sample_name = clarity.get_user_sample_name(sample_name, lenient=True)
+    yield_q30 = clarity.get_expected_yield_for_sample()
```
mwhamgenomics: `get_expected_yield_for_sample` needs a sample ID passed to it. This function should probably be tested as well, since it's patched in the Crawler tests.

Edit: this has actually been fixed, and the PR is not showing them for some reason. dict referencing of `gender aliases` above still stands, everything else is done.

mwhamgenomics (changes requested):

mwhamgenomics (approved):

PR #194, opened 2016-11-03
Evenness

### LINKED ISSUE #191: Calculate Evenness score and upload to Reporting-App
The Evenness score is formally defined in http://www.nature.com/jhg/journal/v61/n7/full/jhg201621a.html?cookies=accepted

In the supplementary document they define Evenness scove E using R code

``` R
C=round(mean(D))
D2=D[D<=C]
E=1-(length(D2)-sum(D2)/C)/length(D)
```

Where D is a vector of number representing the coverage at every bases.
We need to translate that function into python and calculate evenness base on the coverage histogram we obtain from samtools depth.

tcezard:
Add Evenness calculation from the coverage histogram
closes #191


mwhamgenomics:
This looks fine. I'll pull master and accept this

PR #203, opened 2016-11-08
split driver and add freebayes support

### LINKED ISSUE #196: Support multiple Analysis type
There is a new requirement that we can use [freebayes](https://github.com/ekg/freebayes) for variant calling instead of GATK.
Implementing that within bcbio is straightforward as it is just a simple config change but since we also need to support it outside of bcbio we should think of a way of supporting other (potentially multiple per sample) analysis types

tcezard:
close #196


analysis_driver/pipelines/bcbio_pipelines.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+def _run_bcbio(sample_id, sample_dir, sample_fastqs, genome_version, analysis_type):
+    if not genome_version:
+        genome_version = cfg['genome']
+    if not analysis_type:
+        analysis_type = 'gatk'
+    elif analysis_type.endwith('gatk'):
+        analysis_type = 'gatk'
+    elif analysis_type.endwith('freebayes'):
```
mwhamgenomics: This should be `endswith`, not `endwith`

mwhamgenomics (changes requested):

mwhamgenomics (approved):

PR #204, opened 2016-11-08
Pipeline stopping

mwhamgenomics:
The pipeline now reacts to `10` and `15` termination signals, and cleanly sends notifications and stops any running cluster jobs. Note that other signals are not caught.

A running pipeline can be sent a sigterm either by pid via `kill <signal>`, or by dataset name via `analysis_driver --stop <dataset_names>`.

### DIFF
--- a/analysis_driver/client.py
+++ b/analysis_driver/client.py
@@ -49,7 +49,7 @@ def main():
         for d in args.stop:
             pid = scanner.get_dataset(d).most_recent_proc.get('pid')
             if pid:
-                os.kill(pid, 15)
+                os.kill(pid, 10)
 
         if args.report:
             scanner.report()
@@ -106,12 +106,8 @@ def _process_dataset(d):
     )
 
     def _handle_exception(exception):
-        import traceback
-        stacktrace = traceback.format_exc()
         app_logger.critical('Encountered a %s exception: %s', exception.__class__.__name__, str(exception))
-        app_logger.info('Stack trace below:\n' + stacktrace)
         _handle_termination(9)
-        ntf.crash_report(stacktrace)
 
     def _sigterm_handler(sig, frame):
         app_logger.info('Received signal %s. Cleaning up running jobs', sig)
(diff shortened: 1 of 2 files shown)


mwhamgenomics:
The pid comes from the reporting app and is not removed in the event of an uncaught exception, so could be left out of date. At the moment in `analysis_driver --stop`, there's no checking of the pid to make sure it's a valid pipeline process. Could it be possible that we end up stopping a different pid from what we intended (pipelines, other processes, etc)?

analysis_driver/client.py
```diff
@@ -41,6 +46,10 @@ def main():
             scanner.get_dataset(d).reset()
         for d in args.force:
             scanner.get_dataset(d).force()
+        for d in args.stop:
+            pid = scanner.get_dataset(d).most_recent_proc.get('pid')
+            if pid:
+                os.kill(pid, 15)
```
tcezard: Can we change the signal to 10 (USR1) to have something specific to analysis driver so we can distinguish between process killed by analysis driver or by something else

analysis_driver/client.py
```diff
@@ -98,6 +105,24 @@ def _process_dataset(d):
...
+    def _sigterm_handler(sig, frame):
+        app_logger.info('Received signal %s. Cleaning up running jobs', sig)
+        _handle_termination(sig)
+
+    def _handle_termination(_exit_status):
+        stop_running_jobs()
+        d.fail(_exit_status)
+        sys.exit(_exit_status)
+
+    signal.signal(15, _sigterm_handler)
```
tcezard, line 126: Can you also trap signal 10 with the same handler?

analysis_driver/client.py
```diff
@@ -98,6 +105,24 @@ def _process_dataset(d):
         (AsanaNotification, d, cfg.query('notification', 'asana'))
     )
 
+    def _handle_exception(exception):
+        import traceback
+        stacktrace = traceback.format_exc()
+        app_logger.critical('Encountered a %s exception: %s', exception.__class__.__name__, str(exception))
+        app_logger.info('Stack trace below:\n' + stacktrace)
+        _handle_termination(9)
+        ntf.crash_report(stacktrace)
```
tcezard: This line will never be reached as _handle_termination() call sys.exit().

tcezard: We should call crash_report in _handle_termination.
We need the the notification to be triggered (Asana or email) and the stacktrace will be useful to know when/where the process was stopped.

tcezard (changes requested):

analysis_driver/client.py
mwhamgenomics: It looks like there's only a stacktrace to log if there's a `KeyboardInterrupt` (signal 2).

`_handle_termination` will call notifications via `dataset.fail` with the signal number as the exit status. Is this enough information, or would we like an explicit `dataset.sigterm` (or whatever)?

tcezard (approved):

PR #206, opened 2016-11-24
Calculate relatedness

katieemelianova:


analysis_driver/quality_control/calculate_relatedness.py
```diff
@@ -0,0 +1,68 @@
...
+    def get_gatk_genotype_gvcfs_command(self):
+        gVCF_variants = " ". join(["--variant " + i for i in self.gVCF_files])
+        out_prefix = self.project_id + '_genotype_gvcfs'
+        cmd = 'java -jar %s -T GenotypeGVCFs -R %s %s -o %s' % (cfg['tools']['gatk'], self.reference, gVCF_variants, out_prefix)
+        gatk_outfile = out_prefix + '.vcf'
+        return cmd, gatk_outfile
+
+    def get_vcftools_relatedness_command(self, vcftools_input_file):
+        cmd = '%s --relatedness --vcf %s' % (cfg['tools']['vcftools'], vcftools_input_file)
+        vcftools_relatedness_outfile = vcftools_input_file.split('.')[0] + '.relatedness2'
```
tcezard: I seem to recall that this should be ```--relatedness2``` to use the Kinship calculation. This might have an effect on the output file format and might change your parser

analysis_driver/quality_control/calculate_relatedness.py
```diff
@@ -0,0 +1,68 @@
...
+        super().__init__(dataset, working_dir)
+        self.gVCF_files = gVCF_files
+        self.reference = reference
+        self.project_id = project_id
+
+    def get_gatk_genotype_gvcfs_command(self):
+        gVCF_variants = " ". join(["--variant " + i for i in self.gVCF_files])
+        out_prefix = self.project_id + '_genotype_gvcfs'
+        cmd = 'java -jar %s -T GenotypeGVCFs -R %s %s -o %s' % (cfg['tools']['gatk'], self.reference, gVCF_variants, out_prefix)
+        gatk_outfile = out_prefix + '.vcf'
```
tcezard: Can you multithread the GATK call with ```-nt```. set it as a parameter in this function so it is clear when setting the nb cpu used by slurm why you need several.
I would suggest 12 threads and 12 CPU

analysis_driver/quality_control/calculate_relatedness.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def run_vcftools(self, vcftools_input_file):
+        vcftools_relatedness_command, vcftools_outfile = self.get_vcftools_relatedness_command(vcftools_input_file)
+        self.dataset.start_stage('vcftools_relatedness')
+        vcftools_relatedness_executor = executor.execute(
+            vcftools_relatedness_command,
+            job_name='vcftools_relatedness',
+            working_dir=self.working_dir,
+            cpus=2,
+            mem=10
```
tcezard: This is single threaded and does not need 2 cpus

bin/run_qc.py
```diff
@@ -65,6 +65,16 @@ def _parse_args():
     contam_blast.set_defaults(func=contamination_blast)
     return parser.parse_args()
```
tcezard: That return prevents the relatedness parser to be read

tcezard (changes requested):

analysis_driver/pipelines/projects.py
```diff
@@ -3,16 +3,28 @@
...
 
 def project_pipeline(dataset):
-    species = clarity.get_species_from_sample(dataset.name)
+
     project_id = dataset.name
+    samples_for_project = rest_communication.get_documents('samples', where={'project_id': project_id})
+    species_in_project = []
+    for sample in samples_for_project:
+        sample_id = sample.get('sample_id')
+        species_in_lims = clarity.get_species_from_sample(sample_id)
```
tcezard: Species is stored in our DB in the sample schema. you could access it by saying
```python
species = sample.get('species_name')
if not species:
    species = clarity.get_species_from_sample(sample.get('sample_id'))
species_in_project.append(species)
```

analysis_driver/pipelines/projects.py
```diff
@@ -1,9 +1,34 @@
...
+from analysis_driver.quality_control import Relatedness
+from analysis_driver.exceptions import PipelineError
 
 def project_pipeline(dataset):
+
     project_id = dataset.name
-    project = rest_communication.get_document('projects', where={'project_id': project_id})
-    # run the project pipeline
-    exit_status = 0
+    samples_for_project = rest_communication.get_documents('samples', where={'project_id': project_id})
```
tcezard: We should limit this query to the samples that have status "finished".

analysis_driver/pipelines/projects.py
```diff
@@ -1,9 +1,34 @@
...
+        species_in_lims = clarity.get_species_from_sample(sample_id)
+        species_in_project.append(species_in_lims)
+    if len(set(species_in_project)) != 1:
+        raise PipelineError('Wrong number of species in this project: expected 1')
+    species = list(set(species_in_project))[0]
+
+    working_dir = os.path.join(cfg['jobs_dir'], project_id)
+    reference = cfg['references'][species]['fasta']
+    project_source = os.path.join(cfg.query('sample', 'delivery_source'), project_id)
+    gvcf_paths = project_source + '/*/*g.vcf.gz'
```
tcezard: We take the risk of getting a gvcf that we do not want. We should search for the gvcf of each sample we have in ```samples_for_project```

tcezard (changes requested):

analysis_driver/pipelines/projects.py
mwhamgenomics: Something like:

```python
from egcg_core.util import find_file
gvcf_files = []
for s in samples_for_project:
    f = util.find_file(project_source, s['sample_id'], s['user_sample_id'] + '.g.vcf.gz')
    if f:
        gvcf_files.append(f)
```

tcezard (approved):

analysis_driver/pipelines/projects.py
```diff
@@ -1,9 +1,39 @@
...
+from analysis_driver.quality_control import Relatedness
+from analysis_driver.exceptions import PipelineError
 
 def project_pipeline(dataset):
+
     project_id = dataset.name
-    project = rest_communication.get_document('projects', where={'project_id': project_id})
-    # run the project pipeline
-    exit_status = 0
+    samples_for_project = rest_communication.get_documents('samples', where={'project_id': project_id, 'status': 'finished'})
```
katieemelianova: change to aggregate/sample, change endpoint in reporting app

PR #212, opened 2016-12-14
Ete catch unknown taxids

katieemelianova:
When a taxid is not found in the ETE taxid db, make that rank 'Unavailable' rather than breaking

### DIFF
--- a/analysis_driver/quality_control/contamination_blast.py
+++ b/analysis_driver/quality_control/contamination_blast.py
@@ -6,8 +6,7 @@ from egcg_core import executor
 from .quality_control_base import QualityControl
 from analysis_driver.config import default as cfg
 from analysis_driver.exceptions import AnalysisDriverError
-from egcg_core.app_logging import logging_default as log_cfg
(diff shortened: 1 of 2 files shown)


analysis_driver/quality_control/contamination_blast.py
```diff
@@ -59,8 +61,12 @@ def ncbi(self):
...
         '''retrieve the rank of each of the taxa from that taxid's lineage'''
         if not taxon == 'N/A':
-            l = self.ncbi.get_lineage(int(taxon))
-            rank = self.ncbi.get_rank(l)
+            try:
+                l = self.ncbi.get_lineage(int(taxon))
+                rank = self.ncbi.get_rank(l)
+            except ValueError:
+                rank = {0: 'rank unavailable'}
+                app_logger.warning('The taxid %s does not exist in the ETE TAXDB' % (taxon))
```
mwhamgenomics: ContaminationBlast is descended from AppLogger, so this could just be `self.warning` and you wouldn't need to import `log_cfg` above

PR #214, opened 2017-01-17
Stages endpoint

mwhamgenomics:
This accommodates the changes made in EdinburghGenomics/Reporting-App#72. Also contains refactoring and updates to Readme.md.


analysis_driver/dataset.py
```diff
@@ -36,7 +36,12 @@ def dataset_status(self):
 
     @property
     def stages(self):
-        return [s['stage_name'] for s in self.most_recent_proc.get('stages', []) if 'date_finished' not in s]
+        stages = rest_communication.get_documents(
+            'analysis_driver_stages',
+            all_pages=True,
+            where={'date_finished': None}
+        )
```
tcezard, line 43: Shouldn't this query be limited by the ```self.most_recent_proc.proc_id ```?

analysis_driver/dataset.py
```diff
@@ -96,13 +101,13 @@ def _is_ready(self):
...
+        s = self.name
         pid = self.most_recent_proc.get('pid')
         if pid:
-            out.append('(%s)' % pid)
+            s += ' (%s)' % pid
         if self.stages:
-            out.append('-- %s' % (', '.join(self.stages)))
-        return ' '.join(out)
+            s += ' -- ' + ', '.join(self.stages)
+        return s
```
tcezard, line 111: We should cache ```self.stages``` here to avoid querying the REST API twice
```python
stage = self.stages
if stages:
    s += ' -- ' + ', '.join(stages)
```

tcezard (changes requested):

tcezard (approved):

PR #216, opened 2017-01-26
Luigi task runner

mwhamgenomics:
Pipelines are now run through [Luigi](http://luigi.readthedocs.io). This provides infrastructure for segmenting our pipelines into stages, allowing us to run parallel tasks more cleanly. This should also allow us to do complex workflows, e.g. variant calling outside of tools like BCBio, giving us more control over how our BAMs and VCFs are generated.

(At the moment, each pipeline is run as one large stage, but should be split into multiple stages later.)

Also fixing NoCommunicationDataset string serialisation, and lazy-loading Dataset.ntf.


tcezard (approved):
This looks good

PR #218, opened 2017-02-07
BCL validation

mwhamgenomics:
This adds a function that scans an input dir for bcl files, runs `gzip -t` and raises an error if any bcl is invalid gzip. A record of checked files is written, so it will only check each file once.

The function is not currently called, but will be used in future development where it will be run repeatedly as the bcls are generated.


analysis_driver/util/__init__.py
```diff
@@ -45,3 +48,40 @@ def _write_bcbio_csv(run_dir, sample_id, fastqs, user_sample_id):
...
+
+    for t in tile_ids:
+        lane = t[2]  # s_1_1101.bcl.gz
+        lane_dir = 'L00' + lane
+        all_bcls.extend(join(lane_dir, 'C' + str(c + 1) + '.1', t + '.bcl.gz') for c in range(cycles))
+
+    files_to_check = [f for f in all_bcls if f not in validated_bcls]
+
+    app_logger.info('Checking %s files', len(files_to_check))
+    e = ArrayExecutor(['gzip -t %s' % join(basecalls_dir, f) for f in files_to_check], stream=True)
```
tcezard: Are we talking about running up to 231,936 process in parallel on the login node?

mwhamgenomics: In production it would be staggered as the bcls are generated, but it could be a lot of parallel processes if for some reason we run it on a complete dataset. Will run it through executor.

mwhamgenomics:
We need to be able to tell which bcls should be there and which ones shouldn't because the sequencer hasn't done enough cycles yet. We can get the number of completed cycles using `illuminate`:

```python
    cycles = set(illuminate.InteropDataset(run_dir).ExtractionMetrics().data['cycle'])
```

This should be fairly simple, but takes ~0.5 min to run so would not be useful in EdinburghGenomics/Reporting-App#69.

analysis_driver/util/__init__.py
```diff
@@ -45,3 +49,60 @@ def _write_bcbio_csv(run_dir, sample_id, fastqs, user_sample_id):
...
+            'Number of cycles (%s) disagrees with RunInfo (%s)' % (all_cycles[-1], run_info_ncycles)
+        )
+
+    last_completed_cycle = None
+    for c in uniq_cycles:
+        if all_cycles.count(c) == len(tile_ids):
+            last_completed_cycle = c
+            break
+    if last_completed_cycle is None:
+        raise AnalysisDriverError('Bad cycles')
```
tcezard: if `last_completed_cycle` is `None` then it means that the first cycle is not finished isn't it ?
We risk of failing the process because we're too early in the process
maybe set `last_completed_cycle = 0`

mwhamgenomics:
The `check_bcl_files` function is now a class. For example:

```python
run_info = RunInfo(run_dir)
v = BCLValidator(run_dir, run_info, 'path/to/checked_bcls.csv')

while not dataset.rta_complete():
    bcls = v.get_bcls_to_check()
    v.run_bcl_check(bcls)  # or run_bcl_check_local(bcls)
    sleep(60)

invalid_bcls = v.read_invalid_files()
```

tcezard (approved):

PR #220, opened 2017-02-14
Add archiving

### LINKED ISSUE #219: Data archiving at the end of the processing
Archiving the data stored in runs and project folder is currently the responsibility of Robinhood.
We're in the process of retiring Robinhood so we need to implement the arching ourselves.
The copy of the data to tape is actually handled by DMF  so the only thing we need to run is the registration of each file to DMF using:
```lfs hsm_archive```

tcezard:
delete previous analysis of the data before moving
Add archive step
closes #219


tcezard:
This requires EGCG-Core v0.6.6 which need to be tagged

mwhamgenomics (approved):

PR #221, opened 2017-02-17
Multiple genomes per species

katieemelianova:
Support multiple genomes, using the 'Genome Version' udf from the Lims.
Due to this udf being used before to check for BCBIO genome version, the BCBIO genome version now automatically defaults to the genome set in the config.


analysis_driver/pipelines/__init__.py
```diff
@@ -19,15 +19,23 @@ def pipeline(dataset):
...
             raise PipelineError('No species information found in the LIMS for ' + dataset.name)
-        elif species == 'Homo sapiens':
-            return bcbio_var_calling_pipeline(dataset, genome_version, analysis_type)
-        elif clarity.get_sample(dataset.name).udf.get('Analysis Type') == 'Variant Calling':
-            return var_calling_pipeline(dataset, species)
+        if genome_version is None:
+            genome_version = cfg.query('species', species, 'default')
+            reference = cfg.query('genomes', genome_version, 'default')
+        else:
+            reference = cfg.query('genomes', genome_version, 'fasta')
```
tcezard: This should be:
```python
if genome_version is None:
    genome_version = cfg.query('species', species, 'default')
reference = cfg.query('genomes', genome_version, 'fasta')
```
since there isn't any `default` in `genome_version` anymore

analysis_driver/pipelines/__init__.py
```diff
@@ -19,15 +19,23 @@ def pipeline(dataset):
...
+        if species == 'Homo sapiens':
+            return bcbio_var_calling_pipeline(dataset, bcbio_genome, analysis_type)
```
tcezard: bcbio's `genome_version` won't be working anymore as it will force to use `bcbio_genome`

analysis_driver/pipelines/qc_pipelines.py
```diff
@@ -22,7 +22,6 @@ def bam_file_production(dataset, species):
     app_logger.info('Job dir: ' + sample_dir)
     user_sample_id = clarity.get_user_sample_name(sample_id, lenient=True)
 
-    reference = cfg.query('references', species, 'fasta')
     if not reference:
         raise PipelineError('Could not find reference for species %s in sample %s ' % (species, sample_id))
```
tcezard: We should move this exception to when we look for the version

tcezard (changes requested):
`pipeline` should be cleanned a bit to only be resonsible for launching the expected pipeline.
It would be cleaner to have the genome_version selection as a function in common (including raising  exceptions) and call it in the specific pipeline when needed.

tcezard (approved):

analysis_driver/pipelines/common.py
```diff
@@ -69,6 +70,24 @@ def bcbio_prepare_sample(job_dir, sample_id, fastq_files):
...
+        raise PipelineError('Could not find reference for species %s in sample %s ' % (species, dataset_name))
+    return genome_version, reference
+
+
+def get_dbsnp(genome_version):
+    return cfg.query('genomes', genome_version, 'dbsnp')
+
+
+def get_known_indels(genome_version):
+    return cfg.query('genomes', genome_version, 'known_indels')
```
mwhamgenomics, line 88: Do we really need `get_dbsnp` and `get_known_indels`? They don't really do much themselves and are only called once each in variant_calling

tcezard: I think that's fine.
They might be used more once we start the segmentation

PR #224, opened 2017-02-22
Contamination blast error reporting

katieemelianova:
Issue #223 was put up because, even though the contamination check step failed (due to a suspected bus error), the non-zero exit status was not passed on to the main Analysis Driver exit status, thereby not breaking the pipeline and the sample finished with an exit status of 0. 

This pull request should fix this; if contamination check returns a non-zero exit status the sample will fail.


mwhamgenomics (changes requested):
The exit status still needs to be propagated to the final pipeline exit status in the driver script - at the moment, it's only propagated to the stage end

mwhamgenomics:
There isn't any actual reporting of errors done here - are we wanting to do any of that (e.g. raise PipelineError if exit_status != 0)?

Also, handling of exit statuses and file paths is getting a bit messy - should refactor later.

katieemelianova:
Sorry I had a bit of a git glitch, these changes had been added yesterday but not included.

mwhamgenomics (approved):

PR #225, opened 2017-02-22
Generate samtools stats instead of bamtools stats 

### LINKED ISSUE #201: Run Samtools stats
We should run samtools stats instead of bamtools stats and change the parsers accordingly.

tcezard:
closes #201

### DIFF
--- a/analysis_driver/util/bash_commands.py
+++ b/analysis_driver/util/bash_commands.py
@@ -118,12 +118,6 @@ def bwa_mem_biobambam(fastq_pair, reference, expected_output_bam, read_group=Non
     return cmd
 
 
-def bamtools_stats(bam_file, output_file):
-    bamtools_bin = cfg.query('tools', 'bamtools')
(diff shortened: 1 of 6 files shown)


analysis_driver/util/bash_commands.py
```diff
@@ -124,6 +124,12 @@ def bamtools_stats(bam_file, output_file):
     app_logger.debug('Writing: ' + cmd)
     return cmd
 
+def samtools_stats(bam_file, output_file):
+    samtools_bin = cfg.query('tools', 'samtools')
+    cmd = '%s stats %s > %s' % (samtools_bin, bam_file, output_file)
+    app_logger.debug('Writing: ' + cmd)
+    return cmd
+
```
mwhamgenomics: Are we wanting to keep `bamtools_stats` above? Looks like it's been removed from the rest of the pipeline, including the tests

mwhamgenomics (approved):

PR #230, opened 2017-03-03
Early samplesheet loading

katieemelianova:
Starts the run process as soon as the run is available/has started sequencing
Needs to still be merged later with Murray's bcl validation function


analysis_driver/pipelines/demultiplexing.py
```diff
@@ -49,6 +49,10 @@ def demultiplexing_pipeline(dataset):
     crawler.send_data()
     dataset.end_stage('setup')
 
+    while RunDataset.is_sequencing:
```
tcezard: I think you meant 
```python
while dataset.is_sequencing():
```

analysis_driver/dataset.py
```diff
@@ -162,7 +162,10 @@ def __init__(self, name, path, most_recent_proc=None):
         self.path = path
 
     def _is_ready(self):
-        return os.path.isfile(os.path.join(self.path, 'RTAComplete.txt'))
+        return True
+
+    def is_sequencing(self):
+        return clarity.get_run(self.name).udf.get('Run Status') == 'RunStarted'
```
tcezard: Are we sure the run will exist in the LIMS?
What happen if it doesn't?

tcezard (changes requested):

analysis_driver/dataset.py
```diff
@@ -11,6 +12,7 @@
     DATASET_PROCESSING, DATASET_PROCESSED_SUCCESS, DATASET_PROCESSED_FAIL, DATASET_ABORTED, ELEMENT_RUN_NAME,\
     ELEMENT_NB_Q30_R1_CLEANED, ELEMENT_NB_Q30_R2_CLEANED
 
+app_logger = log_cfg.get_logger('versions')
```
mwhamgenomics: Why the name 'versions'? Also this is only used in RunDataset, which is an AppLogger already, so may as well just do `self.warning` there, and you wouldn't need to import `log_cfg`

analysis_driver/quality_control/bcl_validation.py
```diff
@@ -0,0 +1,118 @@
+import csv
+import time
+from os.path import join, isfile
+import illuminate
+from bitstring import ReadError
+from egcg_core import executor
+from egcg_core.util import str_join
+from egcg_core.app_logging import AppLogger, logging_default as log_cfg
+from analysis_driver.exceptions import AnalysisDriverError
+app_logger = log_cfg.get_logger('util')
```
mwhamgenomics: Again, don't need this app_logger.

analysis_driver/pipelines/demultiplexing.py
```diff
@@ -49,8 +50,13 @@ def demultiplexing_pipeline(dataset):
     crawler.send_data()
     dataset.end_stage('setup')
 
-    # Need to sleep to make sure the LIMS has had enough time to update itself
-    time.sleep(900)
```
mwhamgenomics, line 53: Moving the sleep to BCLValidator has made it a bit implicit - maybe comment there about leaving time for the Lims to update. Also, sleeping for an hour might be a bit much

analysis_driver/pipelines/demultiplexing.py
```diff
@@ -1,15 +1,16 @@
-import time
+import os
```
mwhamgenomics: Don't need this `import os` - join is already imported from os.path

mwhamgenomics (changes requested):

analysis_driver/pipelines/demultiplexing.py
tcezard: After Chatting with Murray we agreed that we should bring down the sleep time to 20min (approx 1 cycle)

analysis_driver/dataset.py
```diff
@@ -162,7 +162,14 @@ def __init__(self, name, path, most_recent_proc=None):
...
     def _is_ready(self):
-        return os.path.isfile(os.path.join(self.path, 'RTAComplete.txt'))
+        return True
+
+    def is_sequencing(self):
+        # Assume the run has started and not finished if the status is 'RunStarted' or if it hasn't yet appeared in the LIMS
+        if not clarity.get_run(self.name):
+            self.warning('Run %s not found in the LIMS', self.name)
+            return True
+        return clarity.get_run(self.name).udf.get('Run Status') == 'RunStarted'
```
tcezard: I just realised that the Run object in the genologics API will get cached and that the UDF `'Run Status'` will never change.
We should explicitly store and update the lims run object and update it at each call of `is_sequencing`
```python
@property
def lims_run(self):
    if not self._lims_run:
        self._lims_run = clarity.get_run(self.name)
    return self._lims_run

def is_sequencing(self):
    if not self.lims_run:
        self.warning('Run %s not found in the LIMS', self.name)
        return True
    # force the update of the LIMS object
    self.lims_run.get(force=True)
    return self.lims_run.udf.get('Run Status') == 'RunStarted'
```

tcezard (changes requested):

tcezard (approved):

analysis_driver/pipelines/demultiplexing.py
```diff
@@ -49,8 +49,13 @@ def demultiplexing_pipeline(dataset):
...
-    # Need to sleep to make sure the LIMS has had enough time to update itself
-    time.sleep(900)
+    validation_log = join(job_dir, 'checked_bcls.csv')
+    b = BCLValidator(job_dir, run_info, validation_log, dataset)
+    b.check_bcls()
+    invalid_bcls = b.read_invalid_files()
+    if invalid_bcls:
+        raise AnalysisDriverError('Some BCL files are corrupted; check %s for details', validation_log)
+
     run_status = clarity.get_run(run_id).udf.get('Run Status')
```
tcezard: we should retrieve the status from the stored Lims object with
```python
if dataset.lims_run.udf.get('Run Status') != 'RunCompleted':
```
also remove the TODO as it is now taken care of

tcezard (changes requested):

tcezard (approved):

mwhamgenomics (approved):

PR #233, opened 2017-03-21
Allow any focal species

### LINKED ISSUE #232: Add support for any focal species in Fastq screen reporting
In the fastq screen parser there is a requirement that the species of interest (focal species) be in the database for any reporting to happen.
We should still report the fastqscreen output as it provides information regardless of the focal species.

katieemelianova:
Do not assume focal species will be given, rather give a warning but return empty string for 'focal_species_unique_mapped' if  no fastqscreen species matches focal species name

Fixes #232


tcezard (changes requested):
The default value for `focal_species_pc_unmapped` should be 100 rather than empty string.
Otherwise this looks good

tcezard (approved):

mwhamgenomics (approved):

PR #234, opened 2017-03-22
Segment demultiplexing

mwhamgenomics:
demultiplexing.py is now split into Luigi tasks. Dataset also has a new method to mark a dataset as `resume`. This behaves in the same way as `reprocess`, except when the pipeline is started, it will not take a new `analysis_driver_proc`. Instead it will use the most recent one, and Luigi will only run tasks for stages that do not have an exit status of 0.

RunDataset also now has `run_info` and `sample_sheet` as properties.


analysis_driver/dataset.py
```diff
@@ -73,6 +74,10 @@ def fail(self, exit_status):
     def abort(self):
         self.most_recent_proc.finish(DATASET_ABORTED)
 
+    def resume(self):
+        self.terminate()
+        self.most_recent_proc.change_status(DATASET_RESUME)
+
```
tcezard, line 79: We should change the terminate method to avoid the race condition where the killed process fail after it has been resumed or reseted  
```python
    def terminate(self):
        pid = self.most_recent_proc.get('pid')
        if pid and self._is_running_pid(pid):
            self.info('Terminating pid %s for %s %s', pid, self.type, self.name)
            os.kill(pid, 10)
        while self._is_running_pid(pid):
            time.sleep(1)


    @staticmethod
    def _is_running_pid(pid):        
        #see http://stackoverflow.com/questions/7653178/wait-until-a-certain-process-knowing-the-pid-end
        try:
            os.kill(pid, 0)
        except OSError as err:
            if err.errno == errno.ESRCH:
                return False
        return True
```

analysis_driver/dataset.py
```diff
@@ -2,14 +2,14 @@
...
+from egcg_core.config import cfg
 from egcg_core.app_logging import AppLogger
 from egcg_core.exceptions import RestCommunicationError
 from analysis_driver.exceptions import AnalysisDriverError
-from egcg_core.constants import DATASET_NEW, DATASET_READY, DATASET_FORCE_READY, DATASET_REPROCESS,\
-    DATASET_PROCESSING, DATASET_PROCESSED_SUCCESS, DATASET_PROCESSED_FAIL, DATASET_ABORTED, ELEMENT_RUN_NAME,\
-    ELEMENT_NB_Q30_R1_CLEANED, ELEMENT_NB_Q30_R2_CLEANED
+from analysis_driver.notification import NotificationCentre
+from analysis_driver import reader
+from egcg_core.constants import *
```
tcezard: That will get us an error in Landscape: add as a comment:
```python
from egcg_core.constants import *  # pylint: disable=unused-import
```

analysis_driver/pipelines/demultiplexing.py
```diff
@@ -1,185 +1,175 @@
...
-        json_file = join(fastq_dir, 'demultiplexing_results.json')
-        crawler.write_json(json_file)
+
+class DemultiplexingStage(segmentation.Stage):
+    @property
+    def fastq_dir(self):
+        return join(self.jobs_dir, 'fastq')
+
+
+class Setup(DemultiplexingStage):
```
tcezard, line 21: This stage should be replaced by BCL validator when merging with #230

analysis_driver/pipelines/demultiplexing.py
```diff
@@ -1,185 +1,175 @@
...
+
+        if exists(conversion_xml) and exists(adapter_trim_file):
+            self.info('Found ConversionStats and AdaptorTrimming. Sending data.')
+            crawler = RunCrawler(
+                self.dataset.name, self.sample_sheet, adapter_trim_file=adapter_trim_file,
+                conversion_xml_file=conversion_xml, run_dir=self.fastq_dir
+            )
+            # TODO: review whether we need this
+            json_file = join(self.fastq_dir, 'demultiplexing_results.json')
+            crawler.write_json(json_file)
```
tcezard: We should drop these 3 lines now the `demultiplexing_results.json` file has never been used

tcezard (changes requested):

analysis_driver/dataset.py
```diff
@@ -85,9 +89,15 @@ def force(self):
 
     def terminate(self):
         pid = self.most_recent_proc.get('pid')
-        if pid and self._is_valid_pid(pid):
-            self.info('Terminating pid %s for %s %s', pid, self.type, self.name)
-            os.kill(pid, 10)
+        self.info('Attempting to terminate pid %s for %s %s', pid, self.type, self.name)
+        if not pid or not self._pid_valid(pid):
+            self.error('Attempted to terminate invalid pid %s', pid)
+            return
```
tcezard: The info and error will generate output for any resume and reset for which the process has already finished. It's not really a problem but will look ugly. Can this be changed to debug ?

katieemelianova (approved):

tcezard (changes requested):

tcezard (approved):

PR #235, opened 2017-03-24
Integration test updates

mwhamgenomics:
- `integration_test.py` now automatically sets up a Docker container of EdinburghGenomics/Reporting-App for each test.
- var_calling and qc are now tested
- each test will report an all assertions failed, not just the first encountered
- `--noemail` argument
- also fixed a bug in `client` (encountered during testing) where `log_cfg`'s level was not set correctly


tcezard (approved):
looks good to me

PR #236, opened 2017-04-06
Get samplesheet from lims

### LINKED ISSUE #228: Sample sheet parsing
In Seqlab 2, some of the sample sheet columns have changed:

- Sample_ID and Sample_Name have swapped roles
- Index and Index2 are no longer capitalised
- Project_Name no longer needs to be converted to Sample_Project

Some of the fuzzy logic for locating columns can probably be removed as well, since our sample sheet format is fairly stable now.

tcezard:
Extract the run elements from the LIMS and store them in the RunDataset.
Generate a new Samplesheet.csv from the run elements rather than using the one from the LIMS
closes #228


analysis_driver/dataset.py
```diff
@@ -191,39 +198,116 @@ def run_info(self):
...
         return True
 
     @property
+    def run_elements(self):
+        if not self._run_elements:
+            self._run_elements = self._run_elements_from_lims()
+        return self._run_elements
+
+    def _run_elements_from_lims(self):
+        from egcg_core import clarity
```
mwhamgenomics: This isn't used, so we should remove it

analysis_driver/dataset_scanner.py
```diff
@@ -100,8 +100,9 @@ def __init__(self, config):
 
     def get_dataset(self, name, most_recent_proc=None):
         dataset_path = os.path.join(self.input_dir, name)
+        print(dataset_path)
```
mwhamgenomics: We should remove this

mwhamgenomics (dismissed):
A couple of unused lines should be removed

katieemelianova (approved):

PR #246, opened 2017-05-04
Detect bad tiles interop

tcezard:
This PR adds a new QC that can detect bad tile by parsing the interop files.
This is meant to be used in addition to tile filtering in fastq filterer
The interop package is tricky to install and has caused issue in travis.
On mac only python 3.6 is supported and on linux I've only manage to make it work on 3.4 and 3.6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ python:
   - "3.4"
   - "3.5"
   - "3.6"
+#Command ran before the install
+before_install:
+  - "pip install pip --upgrade"
 # command to install dependencies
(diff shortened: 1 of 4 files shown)


mwhamgenomics (approved):
Just to reiterate, there is bug in `interop` that will prevent it from running on OSX, due to it referencing a dylib path that doesn't exist - see [here](https://github.com/Illumina/interop/issues/127) for a fix.

tcezard:
This feature will be added in a future PR.
Closing for now

PR #253, opened 2017-05-26
Test runcrawler

### LINKED ISSUE #251: Add test for RunCrawler when rundir is provided
The Run crawler is only partially tested and would benefit from having test that provide the rundir.

katieemelianova:
test for barcode updates where run_dir is provided to RunCrawler
resolves #251


analysis_driver/report_generation/report_crawlers.py
```diff
@@ -238,6 +249,20 @@ def _populate_barcode_info_from_fastq_filterer_files(self, run_dir):
...
+
+
+
+
+
+
+
+
+
+
```
mwhamgenomics: ?

analysis_driver/report_generation/report_crawlers.py
```diff
@@ -211,6 +211,17 @@ def _populate_barcode_info_from_seqtk_fqchk_files(self, run_dir):
...
```
mwhamgenomics: ?

mwhamgenomics (changes requested):
Instead of pushing empty fastqfilterer files to Git, can we patch `_populate_barcode_info_from_fastq_filterer_files` in test_crawlers? Or better still, we could add data to the fastqfilterer files and test that.

mwhamgenomics (approved):

tcezard (approved):

PR #258, opened 2017-05-30
Stage multiprocessing fixes

### LINKED ISSUE #256: Multiprocessing and stage handling
- We're getting unprocessable entities in calls to `MostRecentProc.start_stage`, where a call to GET returns something, causing it to do a PATCH on the stage with `"date_finished":null`, which the schema rejects.
- We're getting a lot of calls to `GET analysis_driver_stages?where={"date_finished":null}` in `Dataset.running_stages`
- We're only seeing one stage at a time in the app stage report. This seems to be caused by calling `start_stage` (which consists of multiple GETs and PATCHes) in concurrent Processes, therefore overwriting the analysis_driver_proc's list of stages

mwhamgenomics:
Concurrent stages have each been overwriting `MostRecentProc.entity` with stage IDs because `MostRecentProc.start_stage` was not entirely locked. We were also finding stages that were failing and re-running over and over, which we found to be a Luigi feature. These have been fixed.

Also separated the `SampleDataset` representation we use in `--report` (which calls the Rest API via `running_stages`) from its `__repr__` (which is called many times by Luigi).

Closes #256.

### DIFF
--- a/analysis_driver/client.py
+++ b/analysis_driver/client.py
@@ -120,12 +120,12 @@ def _process_dataset(d):
         stop_running_jobs()
         d.fail(sig)
         sys.exit(sig)
-    # SIGUSR1 is used by luigi to stop submitting new jobs.
+    # SIGUSR1 is used by Luigi to stop submitting new jobs.
     # make sure here that SIGUSR1 is caught even if luigi is not running.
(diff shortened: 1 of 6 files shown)


tcezard (approved):
This looks good

PR #259, opened 2017-05-31
Trios relatedness

katieemelianova:
Adds function to run peddy in Projects pipeline


analysis_driver/pipelines/projects.py
```diff
@@ -1,14 +1,28 @@
...
+
+    species_in_project = set()
+    for sample in dataset.samples_processed:
+        species = sample.get('species_name')
+        if not species:
+            species = clarity.get_species_from_sample(sample.get('sample_id'))
+        species_in_project.add(species)
+    if len(species_in_project) != 1:
+        raise PipelineError('Unexpected number of species (%s) in this project' % ', '.join(species_in_project))
+    species = species_in_project.pop()
```
mwhamgenomics: This functionality is already in ProjectDataset. Just call `dataset.species` instead.

analysis_driver/quality_control/calculate_relatedness.py
```diff
@@ -1,41 +1,52 @@
...
+        return self.run_gatk()
+
+
+class Relatedness(Stage):
+
+    @property
+    def gatk_outfile(self):
+        return os.path.join(self.job_dir, self.dataset.name + '_genotype_gvcfs.vcf')
+
+    project_id = Parameter()
```
mwhamgenomics: Don't need this Parameter, as this will be the same as `self.dataset.name`

analysis_driver/quality_control/calculate_relatedness.py
```diff
@@ -46,4 +57,105 @@ def run_vcftools(self):
         ).join()
 
     def _run(self):
-        return self.run_gatk() + self.run_vcftools()
+        return self.run_vcftools()
+
+
+class Peddy(Stage):
+    ids = Parameter()
```
mwhamgenomics: This should be a ListParameter

etc/output_files.yaml
```diff
@@ -203,3 +203,6 @@ project_process:
   - location: ['']
     basename: '{project_id}.relatedness2'
 
+  - location: ['']
+    basename: '{project_id}.ped_check.csv'
```
mwhamgenomics: I'm not seeing where this file comes from - is it created implicitly by Peddy?

mwhamgenomics (changes requested):

etc/output_files.yaml
katieemelianova: This is created by Peddy using a prefix (here defined as the project ID)

analysis_driver/pipelines/projects.py
```diff
@@ -1,14 +1,27 @@
...
+
+    species_in_project = set()
+    for sample in dataset.samples_processed:
+        species = sample.get('species_name')
+        if not species:
+            species = clarity.get_species_from_sample(sample.get('sample_id'))
+        species_in_project.add(species)
+    if len(species_in_project) != 1:
+        raise PipelineError('Unexpected number of species (%s) in this project' % ', '.join(species_in_project))
+
```
tcezard: Since you're now using `dataset.species` you do not need the code above.

analysis_driver/pipelines/projects.py
```diff
@@ -1,14 +1,27 @@
...
+    project_source = os.path.join(cfg.query('sample', 'delivery_source'), project_id)
     gvcf_files = []
     for sample in dataset.samples_processed:
         gvcf_file = find_file(project_source, sample['sample_id'], sample['user_sample_id'] + '.g.vcf.gz')
```
tcezard, line 18: You're not adding the `gvcf_file` to `gvcf_files`. You should also check if you found gvcf_file

analysis_driver/quality_control/calculate_relatedness.py
```diff
@@ -1,41 +1,50 @@
...
-
-class Relatedness(Stage):
-    gvcf_files = ListParameter()
+class Genotype_gVCFs(Stage):
+    gVCFs = ListParameter()
     reference = Parameter()
-    project_id = Parameter()
 
     @property
     def gatk_outfile(self):
```
tcezard: `gatk_outfile` should be put to a parent class that `Peddy`, `Relatedness`, and `Genotype_gVCFs` would subclass.

tcezard (changes requested):

analysis_driver/pipelines/projects.py
```diff
@@ -17,13 +30,11 @@ def build_pipeline(dataset):
...
-    relatedness = Relatedness(
-        dataset=dataset,
-        gvcf_files=gvcf_files,
-        reference=cfg['references'][dataset.species]['fasta'],
-        project_id=dataset.name
-    )
-    output = Output(dataset=dataset, previous_stages=[relatedness])
+    reference = cfg['references'][dataset.species]['fasta']
+    genotype_gvcfs = Genotype_gVCFs(dataset=dataset, gVCFs=gvcf_files, reference=reference)
+    relatedness = Relatedness(dataset=dataset, project_id=project_id, previous_stages=[genotype_gvcfs])
```
mwhamgenomics: Since `project_id` has been removed from Relatedness, it should also be removed from this constructor call.

analysis_driver/quality_control/calculate_relatedness.py
```diff
@@ -1,41 +1,50 @@
...
-
     def run_gatk(self):
         return executor.execute(
             self.gatk_genotype_gvcfs_cmd(),
             job_name='gatk_genotype_gvcfs',
             working_dir=self.job_dir,
-            cpus=12,
-            mem=30
+            cpus=30,
+            mem=50
```
tcezard: Is this really necessary? 
Especially since the command you're running only has 12 threads?

analysis_driver/quality_control/calculate_relatedness.py
```diff
@@ -46,4 +55,126 @@ def run_vcftools(self):
...
+        ped_file = self.write_ped_file()
+        peddy_cmd = 'peddy --plot --prefix %s %s %s' % (self.dataset.name, self.gatk_outfile + '.gz', ped_file)
+        return peddy_cmd
+
+    def run_peddy(self):
+        return executor.execute(
+            self.peddy_command,
+            job_name='peddy',
+            working_dir=self.job_dir,
+            cpus=10,
```
tcezard: You are using 10 cpus but don't seem to be using any multithreading in Peddy. Are you sure that's right ?

tcezard (changes requested):

analysis_driver/quality_control/calculate_relatedness.py
mwhamgenomics: Also tabix and bgzip above are being run in a single thread and use very little memory, so don't need 12 cpus and 30GB.

bin/run_qc.py
```diff
@@ -177,5 +183,37 @@ def detect_bad_cycle_tile_in_run(dataset, args):
...
+        return gvcfs
+
+def peddy(dataset, args):
+    all_gvcfs = []
+    sample_ids = []
+    if args.samples:
+        cfg.merge(cfg['sample'])
+        for sample_id in args.samples:
+            sample_ids.append(sample_id)
+            s = get_documents(sample_id)[0]
```
mwhamgenomics: Might as well just use `get_document` here, since we're only taking the first element from `get_documents`, and the sample ID is a unique identifier anyway.

bin/run_qc.py
```diff
@@ -177,5 +183,37 @@ def detect_bad_cycle_tile_in_run(dataset, args):
...
+            all_gvcfs.append(gvcf)
+    elif args.projects:
+        cfg.merge(cfg['project'])
+        for project_id in args.projects:
+            samples_for_project = clarity.get_sample_names_from_project(project_id)
+            sample_ids.extend(samples_for_project)
+            project_folder = util.find_file(cfg['input_dir'], project_id)
+            all_gvcfs.extend(get_all_project_gvcfs(project_folder))
+
+    genotyped_gvcfs, gatk_outfile = qc.Genotype_gVCFs(all_gvcfs)
```
mwhamgenomics: This will not work, as it will try to construct a Genotype_gVCFs object with a list of files and unpack that into two variables. Instead, I would build a small pipeline here, and use `all_gvcfs` to construct a Genotype_gVCFs stage object to pass into Peddy via previous_stages.

analysis_driver/pipelines/projects.py
katieemelianova: Sorry, not sure if I've missed something, I believe I'm adding the gvcf_file to gvcf_files on line 29 - are you referring to something else? 

Will address checking the file has been found.

tcezard: Sorry you're right, I missed that. You also do check that the file has been found

analysis_driver/quality_control/calculate_relatedness.py
```diff
@@ -46,4 +53,138 @@ def run_vcftools(self):
...
+                if not all_families.get(family_id):
+                    all_families[family_id] = []
+                all_families[family_id].append(i)
+        return all_families
+
+    @property
+    def ped_file_content(self):
+        ped_file_content = []
+        for family in self.all_families():
+            family_lines = self.get_member_details(family, self.all_families())
```
tcezard: You're calling `self.all_famillies()` twice which means the function will run twice.
You could store it in `ped_file_content` or cache it in `all_families` making it a `@property`

tcezard (approved):

analysis_driver/quality_control/calculate_relatedness.py
```diff
@@ -1,41 +1,48 @@
...
-        )
-
     def run_gatk(self):
         return executor.execute(
             self.gatk_genotype_gvcfs_cmd(),
             job_name='gatk_genotype_gvcfs',
             working_dir=self.job_dir,
             cpus=12,
-            mem=30
+            mem=50
```
mwhamgenomics, line 32: `self.gatk_genotype_gvcfs_cmd` has only been given 20Gb. Can we reduce `mem` here or, if it genuinely needs 50Gb, increase `memory` above?
Edit: fixed.

bin/run_qc.py
```diff
@@ -177,5 +183,37 @@ def detect_bad_cycle_tile_in_run(dataset, args):
...
+        cfg.merge(cfg['project'])
+        for project_id in args.projects:
+            samples_for_project = clarity.get_sample_names_from_project(project_id)
+            sample_ids.extend(samples_for_project)
+            project_folder = util.find_file(cfg['input_dir'], project_id)
+            all_gvcfs.extend(get_all_project_gvcfs(project_folder))
+
+    qc.Genotype_gVCFs(all_gvcfs)
+    p = qc.Peddy(dataset=dataset, ids=sample_ids)
+    p.run()
```
mwhamgenomics: As it stands, Peddy will try to run before the merged gvcf has been generated and crash. You could pass the Genotype_gVCFs object in via `previous_stages`, but would then need to call `luigi.build` as in `pipelines.__init__`. Alternatively, you can run each task manually:

```python
g = qc.Genotype_gVCFs(dataset=dataset, GVCFs=all_gvcfs, reference=args.reference)
g.run()
p = qc.Peddy(dataset=dataset, ids=sample_ids)
p.run()
```

etc/output_files.yaml
```diff
@@ -197,9 +197,14 @@ gatk_var_calling:
...
+        basename: '{project_id}_genotype_gvcfs.vcf.gz'
 
-  - location: ['']
-    basename: '{project_id}.relatedness2'
+    relatedness:
+        location: ['']
+        basename: '{project_id}.relatedness2'
 
+    ped_check:
+        location: ['']
```
mwhamgenomics, line 209: transfer_data defaults the location to the jobs folder, so don't need the `location: ['']`

mwhamgenomics (changes requested):

etc/output_files.yaml
katieemelianova: Currently Output() in projects.py uses location. I'm considering changing this part but for now it is required I think.

mwhamgenomics (approved):

etc/output_files.yaml
mwhamgenomics: This can stay for now, then. Data output's got a bit untidy over time, so can refactor that later.

PR #261, opened 2017-06-15
Fastq filterer fixes

mwhamgenomics:
Fixes a bug in the fastq filterer stage, where sometimes the stats file was not being written. Now, fastq_filterer and pigz are all run in parallel, and the `wait` Bash command is used to wait on all child processes before Slurm exits. Also refactored some of the FastqFilter stage and the generated fastq_filterer.slurm script.


tcezard (approved):
This looks all good

PR #264, opened 2017-06-27
Trios relatedness

katieemelianova:
Parses results of relatedness analysis. Both Peddy and Relatedness are parsed when project process in run. run_qc has an option of parsing one, another or both analyses.


analysis_driver/quality_control/calculate_relatedness.py
```diff
@@ -11,6 +14,111 @@ class RelatednessStage(segmentation.Stage):
...
+    def parse_all(self):
+        exit_status = 0
+        try:
+            peddy_relatedness = self.get_columns(self.peddy_file, ['sample_a', 'sample_b', 'rel'])
+            peddy_vcftools_relatedness = []
+            with open(self.reformat_relatedness_file) as openfile:
+                csvfile = csv.DictReader(openfile)
+                for line in csvfile:
+                    relatedness_samples = (Counter([line['INDV1'], line['INDV2']]))
+                    [peddy_vcftools_relatedness.append([i[0],
```
mwhamgenomics: List comprehension for calling `list.append`? A simple `for` loop would be easier to read

analysis_driver/quality_control/calculate_relatedness.py
```diff
@@ -11,6 +14,111 @@ class RelatednessStage(segmentation.Stage):
...
+
+    @property
+    def reformat_relatedness_file(self):
+        relatedness_reformat = (os.path.join(self.job_dir, self.dataset.name + '.relatedness2_reformat'))
+        with open(os.path.join(self.job_dir, self.dataset.name + '.relatedness2')) as openfile:
+            readfile = openfile.read()
+            readfile_csv = re.sub('\t', ',', readfile)
+            with open(relatedness_reformat, 'w') as outfile:
+                outfile.write(readfile_csv)
+        return relatedness_reformat
```
mwhamgenomics: Don't see why we're replacing `\t` with `,` here - `DictReader` should be able to take tab-separated files

mwhamgenomics (changes requested):
Some logic that I don't think is needed, and can be removed. Also, tests plz.

analysis_driver/quality_control/calculate_relatedness.py
```diff
@@ -1,16 +1,194 @@
...
+            for e in egc_lines:
+                egc_outfile.write('\t'.join(e) + '\n')
+
+
+    def get_outfile_content(self, values):
+        gel_lines = []
+        egc_lines = []
+        user_ids = self.user_sample_ids()
+        for r in values:
+            if r['sample1'] != r['sample2']:
```
mwhamgenomics, line 83: This skipping of lines comparing a sample to itself is not tested.
Edit: done now.

analysis_driver/quality_control/calculate_relatedness.py
```diff
@@ -1,16 +1,194 @@
...
+
+    def get_columns(self, filename, column_headers):
+        columns_to_return = []
+        try:
+            with open(filename) as openfile:
+                csvfile = csv.DictReader(openfile)
+                for line in csvfile:
+                    columns_to_return.append([line[i] for i in column_headers])
+        except FileNotFoundError:
+            raise PipelineError('Could not find file: %s' % filename)
```
mwhamgenomics: Don't really need this, since if you don't catch the error, you'll get...
```python
>>> open('nonexistent_file.txt')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
IOError: [Errno 2] No such file or directory: 'nonexistent_file.txt'
```
... which says exactly the same thing without needing to try/catch anything. There are a couple of other places here where this is also the case.

mwhamgenomics (changes requested):
Still some unneeded logic, and some untested behaviour

analysis_driver/quality_control/calculate_relatedness.py
katieemelianova: I've changed it here, the other instances of this are a little different, where they return a non-zero exit status instead of crashing and prividing a FileNotFoundError - ought these be changed as well?

mwhamgenomics (approved):

PR #269, opened 2017-07-20
Report in LIMS when a pipeline starts or finishes in a sample

katieemelianova:
Fix run_qc bug
update egcg_core to 0.7 in requirements


analysis_driver/dataset.py
```diff
@@ -65,19 +72,23 @@ def start(self):
...
 
     def fail(self, exit_status):
         self._assert_status(DATASET_PROCESSING)
         self.ntf.end_pipeline(exit_status)
+        if self.type == 'sample': self.lims_ntf.fail_and_reset_sample_pipeline()
         self.most_recent_proc.finish(DATASET_PROCESSED_FAIL)
 
     def abort(self):
         self.most_recent_proc.finish(DATASET_ABORTED)
+        if self.type == 'sample': self.lims_ntf.fail_and_reset_sample_pipeline()
```
mwhamgenomics: I'd have these in SampleDataset as additions to the default behaviour:

```python
class SampleDataset(Dataset):
    def abort(self):
        super().abort()
        self.lims_ntf.fail_and_reset_sample_pipeline()
```

analysis_driver/notification.py
```diff
@@ -1,4 +1,59 @@
-from egcg_core import notifications
+from egcg_core import notifications, clarity
+from pyclarity_lims.lims import Step
```
mwhamgenomics: This should be pyclarity_lims.entities. Also, this should be added to requirements.txt

mwhamgenomics (changes requested):

analysis_driver/dataset.py
```diff
@@ -444,6 +440,22 @@ def report(self):
...
+        super().start()
+        self.lims_ntf.start_sample_pipeline()
+
+    def abort(self):
+        super().abort()
+        self.lims_ntf.fail_and_reset_sample_pipeline()
+
+    def succeed(self):
+        super().succeed()
+        self.lims_ntf.fail_and_reset_sample_pipeline()
```
mwhamgenomics: Failing and resetting a successful pipeline?

mwhamgenomics (changes requested):

analysis_driver/dataset.py
katieemelianova: Oops! Thanks for pointing that out, have changed it now

tests/test_dataset.py
```diff
@@ -348,6 +352,41 @@ def setup_dataset(self):
         ]
         self.dataset._data_threshold = 1000000000
 
+    @patched_initialise
+    @patch(ppath + 'MostRecentProc.start')
+    @patch(ppath + 'MostRecentProc.retrieve_entity')
+    @patch(ppath + 'Dataset.lims_ntf', new_callable=PropertyMock)
+    def test_start(self, mocked_lims_ntf, mocked_retrieve_entity, mocked_start, mocked_update):
+        self.dataset.most_recent_proc.entity['status'] = c.DATASET_PROCESSING
```
mwhamgenomics: Don't see why we're setting an item here when we're just `del`-ing it two lines later

mwhamgenomics (changes requested):

tests/test_dataset.py
katieemelianova: I forgot to delete that line, thanks for pointing out

mwhamgenomics (approved):

analysis_driver/notification.py
```diff
@@ -1,4 +1,59 @@
-from egcg_core import notifications
+from egcg_core import notifications, clarity
+from pyclarity_lims.entities import Step
+from analysis_driver.exceptions import PipelineError
+
+
+class LimsNotification():
+    def __init__(self, name):
+        self.name = name
+        self.sample_list = clarity.get_list_of_samples(self.name)
```
tcezard: We should only retrieve one and only one sample here.
`clarity.get_sample()` would be more appropriate and we should check the return is not `None`

analysis_driver/notification.py
```diff
@@ -1,4 +1,59 @@
...
+        self.artifacts = [sample.artifact for sample in self.sample_list]
```
tcezard: Same as above, only one `artifact`

analysis_driver/notification.py
```diff
@@ -1,4 +1,59 @@
...
+        self.step = None
+
+    @property
+    def lims(self):
+        return clarity.connection()
+
+    @property
+    def postseqlab_stages(self):
+        return self.lims.get_workflows(name='PostSeqLab EG 1.0 WF')[0].stages
```
tcezard: Put the workflow name in the config file.

analysis_driver/notification.py
```diff
@@ -1,4 +1,59 @@
...
+        stage = [stage for stage in self.postseqlab_stages if stage == stage_name]
+        if not len(stage) == 1:
+            raise PipelineError('Can not find %s from Workflow Stages' % stage_name)
+        return stage[0]
+
+    def route_sample_to_data_processing(self):
+        self.lims.route_artifacts(self.artifacts, stage_uri=self.get_stage('Data Processing EG 1.0 ST').uri)
+
+    def create_step(self):
+        self.step = Step.create(self.lims, protocol_step=self.get_stage('Data Processing EG 1.0 ST').step, inputs=[self.artifacts], container_type_name='Tube')
```
tcezard: If `self.artifacts` is a list then you would pass a list of of one list to inputs.
Make sure the artifact is properly passed to `route_artifacts` and  `inputs`

analysis_driver/notification.py
```diff
@@ -1,4 +1,59 @@
...
+
+    @property
+    def lims(self):
+        return clarity.connection()
+
+    @property
+    def postseqlab_stages(self):
+        return self.lims.get_workflows(name='PostSeqLab EG 1.0 WF')[0].stages
+
+    def get_stage(self, stage_name):
```
tcezard: The stage is the same across the entire run and does not need to be searched for every time.

analysis_driver/dataset.py
```diff
@@ -31,6 +29,7 @@ def __init__(self, name, most_recent_proc=None):
         self.name = name
         self.most_recent_proc = MostRecentProc(self.type, self.name, most_recent_proc)
         self._ntf = None
+        self._lims_ntf = None
```
tcezard: Does this need to be in `Dataset` if it is only going to be used in the `SampleDataset`

tcezard (changes requested):

tcezard (changes requested):
Also LimsNotification is not tested.
Can you add unit tests

tests/test_dataset.py
```diff
@@ -348,6 +352,32 @@ def setup_dataset(self):
...
         self.dataset._data_threshold = 1000000000
 
+    @patched_initialise
+    @patch(ppath + 'MostRecentProc.start')
+    @patch(ppath + 'MostRecentProc.retrieve_entity')
+    @patch(ppath + 'SampleDataset.lims_ntf', new_callable=PropertyMock)
+    def test_start(self, mocked_lims_ntf, mocked_retrieve_entity, mocked_start, mocked_update):
+        with patched_get_run:
+            self.dataset.start()
+        self.assertTrue(mocked_lims_ntf.called)
```
tcezard: This doesn't test that `start_sample_pipeline` was called

tests/test_dataset.py
```diff
@@ -348,6 +352,32 @@ def setup_dataset(self):
...
+
+    @patched_finish
+    @patch(ppath + 'MostRecentProc.retrieve_entity')
+    @patch(ppath + 'SampleDataset.lims_ntf', new_callable=PropertyMock)
+    def test_succeed(self, mocked_lims_ntf, mocked_retrieve_entity, mocked_finish):
+        with patched_get_run:
+            self.dataset.most_recent_proc.entity['status'] = c.DATASET_PROCESSING
+            self.dataset.succeed()
+        self.assertTrue(mocked_lims_ntf.called)
```
tcezard: This doesn't test that `assign_next_and_advance_step` was called

tcezard (changes requested):

tests/test_dataset.py
```diff
@@ -336,24 +320,6 @@ def test_no_data_threshold(self, mocked_exp_yield):
...
-        exp = {
-            'name': 'qc',
-            'toolset_type': 'non_human_sample_processing',
-            'toolset_version': 0
-        }
-
-        with patched_get_doc(None), patched_patch as mocked_patch:
-            assert self.dataset._pipeline_instruction() == exp
-            mocked_patch.assert_called_with('projects', {'sample_pipeline': exp}, 'project_id', 'a_project')
-
```
mwhamgenomics, line 353: Hey, we needed that!

tests/test_dataset.py
```diff
@@ -179,25 +174,14 @@ def test_raise_exceptions(self):
...
-            'toolset_version': 0
-        }
+
+class _TestDataset(Dataset):
+    type = 'None'
+    endpoint = 'None'
+    id_field = 'None'
+
+    def _is_ready(self):
+        pass
```
mwhamgenomics, line 184: This was removed in master, but seems to have crept back in during the merge. Can this be reverted, along with the removal of `patched_is_ready`?

tests/test_dataset.py
```diff
@@ -179,25 +174,14 @@ def test_raise_exceptions(self):
...
-        assert self.dataset.pipeline.name == 'qc'
-
-    @patch(ppath + 'toolset', new=Toolset())
-    def test_pipeline_instruction(self):
-        self.dataset._default_pipeline = Mock(return_value='qc')
-        assert self.dataset._pipeline_instruction() == {
-            'name': 'qc',
-            'toolset_type': 'non_human_sample_processing',
-            'toolset_version': 0
-        }
```
mwhamgenomics, line 197: Hey, we needed that!

mwhamgenomics (changes requested):
Some changes from master were overwritten. These should be restored.

PR #270, opened 2017-08-02
Tool versioning

### LINKED ISSUE #222: Support for multiple version of tools
We want to be able to update our tools while still guaranteeing that samples from a project will be run with a specific stable set of tools.
We need to be able to associate a set of tools (config file) and a workflow (analysis_driver version) with a project. The config file and analysis_driver version will be retrieved before analysis driver starts.

One option would be to split Analysis-Driver in two: 
 - Process manager on one side
 - Pipeline runner on the other

Another option would be to add a wrapper script around AnalysisDriver to select the version to run.

mwhamgenomics:
Adds a `toolset` object, which contains all versioned and unversioned tools relevant to the pipeline being run. Anything that runs a third party dependency should use `toolset[toolname]`, __not__ `cfg['tools'][toolname]`.

`tool_versioning.yaml` describes all the dependencies that should be in a toolset, and allows the updating of toolsets with new dependency versions.

The MostRecentProc now records the version of a toolset used on a dataset in `analysis_driver_procs`. By default, the latest version of a toolset is used. The situation with samples is a little more complex:
- SampleDatasets check the `projects` endpoint for a toolset version to use
- the first SampleDataset in a project will query this, find nothing and select the latest toolset
- the toolset version used is recorded in `projects`
- subsequent SampleDatasets will query this and select the same toolset version as the first sample
- this way, all samples in a project will be processed with the same toolset

This builds on and replaces the functionality in reader.VersionReader.
Closes #222.


analysis_driver/dataset.py
```diff
@@ -436,6 +482,17 @@ def report(self):
             s += '(no non useable run elements)'
         return s
 
+    def resolve_pipeline(self):
```
tcezard: The resolve pipeline function seems to only find the pipeline based on the type of sample.
Should this be rename as `default_pipeline`?
Another method should retrieve the saved pipeline from the project based on the saved name.

analysis_driver/dataset.py
```diff
@@ -182,6 +189,14 @@ def report(self):
     def __lt__(self, other):
         return self.name < other.name
 
+    def resolve_pipeline(self):
+        raise NotImplementedError
+
+    def resolve_pipeline_and_toolset(self):
```
tcezard: What about something like:
```python
def resolve_pipeline_and_toolset():
    upload = False
    self._pipeline, self._toolset_version = self.retrieve_saved_version()
    if not self._pipeline:
        self._pipeline = self.resolve_pipeline()
        upload = True
    toolset.select_type(self.pipeline.toolset_type)
    if not self._toolset_version:
        self._toolset_version = toolset.latest_version
        upload = True
    toolset.select_version(self._toolset_version)
    if upload:
        self.push_saved_version()
```
`resolve_pipeline`, `retrieve_saved_version` and `push_saved_version` need to be specific to the dataset type.

tcezard (changes requested):

tcezard (approved):

PR #271, opened 2017-08-14
Project process md5

### LINKED ISSUE #267: Project process data output is missing MD5s
`projects.Output` is missing MD5 checking. There's also some code duplication between this and sample output. #262 will make `transfer_data.create_output_links` generic, so should be able to use that.

katieemelianova:
run md5sum after project process
closes #267


etc/example_analysisdriver.yaml
```diff
@@ -83,3 +92,8 @@ default:
 
     well_duplicate:
       coord_file: path/to/coord_file
+
+    references:
+        Homo sapiens:
+            fasta: /lustre/home/production/Resources/genomes/Homo_sapiens/hg38.fa
+            dbsnp: /lustre/home/production/Applications/bcbio/0.9.8/genomes/Hsapiens/hg38/variation/dbsnp-147.vcf.gz
```
mwhamgenomics: The 'references' config field was removed in #221. This should be:
```yaml
species:
    'Homo sapiens':
        default: a_genome

genomes:
    a_genome:
        fasta: path/to/genome.fa
        dbsnp: path/to/dbsnp.vcf.gz
```
`projects.build_pipeline` should then query this via `dataset.reference_genome`

tests/test_pipelines/test_projects.py
```diff
@@ -0,0 +1,73 @@
...
+    def test_run(self, mocked_execute):
+        with patch('analysis_driver.segmentation.BasicStage.job_dir', new=os.path.join(self.assets_path, 'test_projects')):
+            self.md5._run()
+
+
+        mocked_execute.assert_called_once_with(cpus=1,
+                                               job_name='md5sum',
+                                               log_commands=False,
+                                               mem=2,
+                                               working_dir='/Users/katie/PycharmProjects/Analysis-Driver/tests/assets/test_projects')
```
mwhamgenomics: This is breaking the build.

analysis_driver/pipelines/projects.py
```diff
@@ -1,25 +1,29 @@
...
 from analysis_driver.exceptions import PipelineError
-from analysis_driver.transfer_data import output_data_and_archive
+from analysis_driver.transfer_data import output_data_and_archive, create_output_links
 
 toolset_type = 'project_processing'
 name = 'project'
 
 
+def delivery_source():
+    return cfg.query('sample', 'delivery_source')
```
mwhamgenomics: Looks like this is only called in one place - does it really need to be a function?

analysis_driver/pipelines/projects.py
```diff
@@ -28,35 +32,39 @@ def build_pipeline(dataset):
...
-                if os.path.islink(symlink_path):
-                    os.unlink(symlink_path)
-                os.symlink(source, symlink_path)
-            else:
-                raise PipelineError('Could not find the file ' + source + ', unable to create link')
+        create_output_links(
+            self.job_dir,
+            output_cfg,
+            dir_with_linked_files
+        )
```
mwhamgenomics, line 63: Also missing `toolset.write_to_yaml` for dependency version reporting. I'm feeling like version reporting and md5s should be a standard part of any data output, so there should be a generic DataOutput class in `pipelines.common` that puts all this alongside link creation and the call to output_data_and_archive.

mwhamgenomics (changes requested):

analysis_driver/pipelines/projects.py
katieemelianova: I've added toolset.write_to_yaml though have left writing a generic DataOutput for another pull request

mwhamgenomics (approved):

PR #272, opened 2017-08-15
End-to-end test improvements

mwhamgenomics:
The end to end tests are now more flexible and manageable.

- integration_cfg now has more options
  - a specific branch to switch to when setting up Rest API containers (defaults to master)
  - run/sample IDs, barcodes, etc. of test data are now configurable

- more flexible output file checking
  - `md5s` has been renamed to `files`
  - if filename ends in '.gz', this is piped through `zcat` and then `md5sum`
    - this md5s the actual file content, rather than the compressed data
  - otherwise if there is a file called `filename.md5`, then this md5 file is parsed
  - otherwise if the value for filename is `null`, the existence of the file is checked
    - this lets us at least do some kind of testing on non-deterministic files with timestamps, etc.

- qc data in the Rest API can now be queried with dot notation
  - this way, we can check bits of the qc data, even if other bits of it are non deterministic
- each test method now patches `cfg` so the job and output dirs for each test is unique
  - these are cleaned up at the start of each new test run

- added pipeline_used and stages run to the tests
- [__config change__] changed `well_duplicate` to `well_duplicates` because it was annoying me. Can revert if we don't want this done here/now.
- partially fixed a bug where duplicated Handlers were logging messages multiple times after the first test run
- should we rename the `integration_tests` module to something more technically correct?

All of this should allow us to run the tests routinely without too much sifting of log files.


tcezard (approved):

katieemelianova (approved):

integration_tests/example_integration_test.yaml
```diff
@@ -1,26 +1,34 @@
...
-bcbio:
-    md5s:
-        programs.txt: <md5um>
-        sample-gatk-haplotype.vcf.gz: <md5um>
+a_test_method:
+    files:
+        file_1.txt: <md5>  # there should be a file called file_1.txt.md5 with this content
+        file_2.txt: null  # file_2.txt should exist
+        file_3.txt.gz: <md5>  # zcat file_3.txt.gz should give this md5
+        # ...
```
tcezard: Maybe add example with bam file
```yaml
file_4.bam: <md5>  # samtools idxstat file_4.bam | awk '{seq_len+=$2; mapped+=$3; unmapped+=$4}END {print seq_len,mapped,unmapped}' should give this md5

File: integration_tests/example_integration_test.yaml [hunk 2]
```diff
@@ -1,26 +1,34 @@
...
     qc:
-        bam_file_reads: 7000000
-        genome_size: 3000000000
-        duplicate_reads: 700000
-        called_gender: 'unknown'
+        this: 0
+        that.other: 'another'
+        # ...
+        pipeline_used: {'name': 'a_pipeline', 'toolset_version': 0, 'toolset_type': 'a_toolset_type'}
+        stages: ['stage_1', 'stage_2', 'stage_3']
```
tcezard, line 35: order of the stages is not tested

tcezard (commented):
This looks good

integration_tests/example_integration_test.yaml
mwhamgenomics: If two stages start at the same time, their order wouldn't be predictable, so I don't think this can be tested

PR #283, opened 2017-09-28
Fix bugs in project process

katieemelianova:
Fix discrepancy between projects name and keyword during selection of toolset
Fix Incorrect Sex key (Female - > F, Male -> M)


analysis_driver/quality_control/relatedness.py
```diff
@@ -88,19 +95,20 @@ def get_outfile_content(self, values):
 
                 if len(list(comparison['family_ids'])) == 1:
                     if len(comparison['proband']) == 1:
-                        gel_line = [''.join(list(comparison['family_ids'])),
-                                    comparison['proband'][0],
+                        gel_line = [comparison['proband'][0],
+                                    ''.join(list(comparison['family_ids'])),
                                     'Proband',
                                     comparison['other'][0],
+                                    ''.join(list(comparison['family_ids'])),
```
mwhamgenomics: Duplicated field?

katieemelianova: This is a little bit superfluous but clarifies that the two individuals being compared are from the same family.

analysis_driver/quality_control/relatedness.py
```diff
@@ -52,16 +56,32 @@ def relatedness_file(self):
         return os.path.join(self.job_dir, self.dataset.name + '.relatedness2')
 
     def write_results(self, gel_lines, egc_lines):
-        header = '\t'.join(['S1', 'S1 Family ID', 'S1 Relationship', 'S2', 'S2 Family ID', 'S2 Relationship'])
+        gel_header = '\t'.join(['S1',
```
tcezard: The header order for this family check file should be S1 Family ID
```python
['Family ID', 'S1', 'S1 Relationship', 'S2', 'S2 Relationship', 'Relatedness']
```

tcezard (changes requested):

analysis_driver/quality_control/relatedness.py
katieemelianova: Will also change internal sample ID to user sample ID in output file (making a note of it so I don't forget)

tcezard (approved):

PR #288, opened 2017-11-08
Align run elements

### LINKED ISSUE #181: Calculate early estimate of duplicate and mapping rate at run
There is a growing need for early assessment of some metrics through alignment at the run level.
This assessment needs to be as fast as possible and still fairly accurate.
One option is to explore the possibility of mapping all the reads to the target genome.
This would give us an early estimate of:
- mapping rate
- well duplicates
- we could extrapolate the coverage assuming there isn't much more library duplicates.

We should first test this method before implementing
(See CR359)

tcezard:
Add new steps in demultiplexing pipeline performing alignment and coverage calculation.
Upload the new metrics to the reporting app.
close #181


mwhamgenomics (approved):
Looks fine, but Stages can definitely be made more flexible so they can run repeatedly. Will open an issue for that.

katieemelianova (approved):

PR #296, opened 2017-11-21
Fix run element mapping

### LINKED ISSUE #295: Run element mapping crash because the picard temp dir is not set
It should just be a matter of setting it with `-Djava.io.tmpdir`

tcezard:
close #295

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Changelog for Analysis-Driver
 0.19 (2017-11-16)
 -----------------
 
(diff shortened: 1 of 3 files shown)


mwhamgenomics (approved):

katieemelianova (approved):

tcezard:
Integration tests in `logs/integration_tests/2017-11-21_11\:48\:23.log ` run agains the full size dataset shows that the pipeline finished successfully. Although the rest of the test failed because the dataset's metric and md5 did not match the yaml file

Issue #286, opened 2017-10-31
Sample processing should use yield instead of yield Q30


tcezard:
To be consistent with the rest of the review process the requirement for a sample to be ready will be changed to: 
 - Yield useable > Require yield
 - %Q30 > 75%

tcezard:
associated with CR572

PR #303, opened 2017-12-12
Sample ready requirements

katieemelianova:
require %q30 to be > 75 for sample to be ready
Update tests

### DIFF
--- a/analysis_driver/dataset.py
+++ b/analysis_driver/dataset.py
@@ -415,6 +415,7 @@ class SampleDataset(Dataset):
     def __init__(self, name, most_recent_proc=None, data_threshold=None):
         super().__init__(name, most_recent_proc)
         self._run_elements = None
+        self._sample = None
         self._non_useable_run_elements = None
         self._data_threshold = data_threshold
         self._species = None
@@ -459,6 +460,14 @@ class SampleDataset(Dataset):
         return self._run_elements
 
     @property
+    def sample(self):
+        if self._sample is None:
+            self._sample = rest_communication.get_document(
+                'aggregate/samples', match={'sample_id': self.name}
+            )
+        return self._sample
+
+    @property
     def non_useable_run_elements(self):
         if self._non_useable_run_elements is None:
             self._non_useable_run_elements = rest_communication.get_documents(
@@ -471,12 +480,7 @@ class SampleDataset(Dataset):
         return self.run_elements[0]['project_id']
 
     def _amount_data(self):
-        return sum(
-            [
-                int(r.get(ELEMENT_NB_Q30_R1_CLEANED, 0)) + int(r.get(ELEMENT_NB_Q30_R2_CLEANED, 0))
-                for r in self.run_elements
-            ]
-        )
+        return self._sample.get('clean_yield_q30')
 
     @property
     def pc_q30(self):
@@ -489,7 +493,7 @@ class SampleDataset(Dataset):
         if self._data_threshold is None:
             self._data_threshold = rest_communication.get_document('samples', where={'sample_id': self.name}).get('required_yield')
         if not self._data_threshold:
-            raise AnalysisDriverError('Could not find data threshold in LIMS for ' + self.name)
+            raise AnalysisDriverError('Could not find data threshold for ' + self.name)
         return self._data_threshold
 
     def _pipeline_instruction(self):
--- a/analysis_driver/dataset.py
+++ b/analysis_driver/dataset.py
@@ -479,6 +479,12 @@ class SampleDataset(Dataset):
         )
 
     @property
+    def pc_Q30(self):
+        total_reads_q30 = sum([r.get(ELEMENT_NB_Q30_R1) + r.get(ELEMENT_NB_Q30_R2) for r in self.run_elements])
+        total_number_of_reads = sum([r.get(ELEMENT_NB_BASE_R1) + r.get(ELEMENT_NB_BASE_R2) for r in self.run_elements])
+        return (total_reads_q30/total_number_of_reads) * 100
+
+    @property
     def data_threshold(self):
         if self._data_threshold is None:
             self._data_threshold = clarity.get_expected_yield_for_sample(self.name)
@@ -498,7 +504,7 @@ class SampleDataset(Dataset):
         return instruction
 
     def _is_ready(self):
-        return self.data_threshold and int(self._amount_data()) > int(self.data_threshold)
+        return self.data_threshold and self.pc_Q30 > 75 and int(self._amount_data()) > int(self.data_threshold)
 
     def report(self):
         runs = sorted(set(r.get(ELEMENT_RUN_NAME) for r in self.run_elements))
--- a/analysis_driver/dataset.py
+++ b/analysis_driver/dataset.py
@@ -495,7 +495,7 @@ class SampleDataset(Dataset):
         return instruction
 
     def _is_ready(self):
-        return self.data_threshold and self.pc_q30 > 75 and int(self._amount_data()) > int(self.data_threshold)
+        return self.data_threshold and self.pc_q30 > 75 and int(self._amount_data()) > self.data_threshold
 
     def report(self):
         runs = self.sample.get('aggregated').get('run_ids')
@@ -509,7 +509,6 @@ class SampleDataset(Dataset):
         else:
             s += '(no non useable run elements)'
         return s
-
 
     def start(self):
         super().start()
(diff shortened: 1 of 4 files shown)


analysis_driver/dataset.py
```diff
@@ -478,6 +478,12 @@ def _amount_data(self):
             ]
         )
 
+    @property
+    def pc_Q30(self):
```
mwhamgenomics: PEP8, plz (lower case)

mwhamgenomics (changes requested):
What issue does this address? If it's #286, there's another part that still needs done.

Also, do we want to switch over to the new aggregation as well, since we're here?

analysis_driver/dataset.py
katieemelianova: Will do, thanks! Aware of the second part of this, will be added in a separate commit.

analysis_driver/dataset.py
```diff
@@ -458,6 +459,14 @@ def run_elements(self):
             )
         return self._run_elements
 
+    @property
+    def sample(self):
+        if self._sample is None:
+            self._sample = rest_communication.get_document(
+                'aggregate/samples', match={'sample_id': self.name}
```
mwhamgenomics: We should use the new aggregation now in the standard endpoints, e.g. `get_document('samples', where={'sample_id': self.name})['aggregated']`

analysis_driver/dataset.py
```diff
@@ -458,6 +459,14 @@ def run_elements(self):
             )
         return self._run_elements
```
mwhamgenomics, line 460: With the new aggregation, we should be able to remove self.run_elements and only use self.sample

analysis_driver/dataset.py
```diff
@@ -471,19 +480,20 @@ def project_id(self):
...
+    def pc_q30(self):
+        total_reads_q30 = sum([r.get(ELEMENT_NB_Q30_R1) + r.get(ELEMENT_NB_Q30_R2) for r in self.run_elements])
+        total_number_of_reads = sum([r.get(ELEMENT_NB_BASE_R1) + r.get(ELEMENT_NB_BASE_R2) for r in self.run_elements])
+        return (total_reads_q30/total_number_of_reads) * 100
 
     @property
     def data_threshold(self):
         if self._data_threshold is None:
-            self._data_threshold = clarity.get_expected_yield_for_sample(self.name)
+            self._data_threshold = rest_communication.get_document('samples', where={'sample_id': self.name}).get('required_yield')
```
mwhamgenomics: This should use `self.sample`.

mwhamgenomics (changes requested):

analysis_driver/dataset.py
```diff
@@ -451,20 +450,10 @@ def user_sample_id(self):
...
-            )
-        return self._run_elements
-
     @property
     def sample(self):
         if self._sample is None:
-            self._sample = rest_communication.get_document(
-                'aggregate/samples', match={'sample_id': self.name}
-            )
+            self._sample = rest_communication.get_document('samples', match={'sample_id': self.name})
```
mwhamgenomics: This should be `where={'sample_id': self.name}`

mwhamgenomics (changes requested):
`SampleDataset._amount_data` now returns a value in Gb and is compared with `required_yield` which is in bases, so will never be over the threshold

mwhamgenomics (approved):

analysis_driver/dataset.py
```diff
@@ -468,22 +466,24 @@ def non_useable_run_elements(self):
...
-            [
-                int(r.get(ELEMENT_NB_Q30_R1_CLEANED, 0)) + int(r.get(ELEMENT_NB_Q30_R2_CLEANED, 0))
-                for r in self.run_elements
-            ]
-        )
+        return int(self._sample['aggregated'].get('clean_yield_q30_in_gb')*1000000000)
+
+    @property
+    def pc_q30(self):
+        s = self.sample
```
tcezard: This should simply be `return self.sample.get('aggregated').get('pc_q30')`

analysis_driver/dataset.py
```diff
@@ -468,22 +466,24 @@ def non_useable_run_elements(self):
...
+        return self.sample['project_id']
 
     def _amount_data(self):
-        return sum(
-            [
-                int(r.get(ELEMENT_NB_Q30_R1_CLEANED, 0)) + int(r.get(ELEMENT_NB_Q30_R2_CLEANED, 0))
-                for r in self.run_elements
-            ]
-        )
+        return int(self._sample['aggregated'].get('clean_yield_q30_in_gb')*1000000000)
```
tcezard: why are you using `self._sample` instead of self.`sample` also if we're switching to yield instead of yield Q30 it should be  `clean_yield_in_gb`

analysis_driver/dataset.py
```diff
@@ -468,22 +466,24 @@ def non_useable_run_elements(self):
...
+
+    @property
+    def pc_q30(self):
+        s = self.sample
+        reads_q30 = s.get(ELEMENT_NB_Q30_R1) + s.get(ELEMENT_NB_Q30_R2)
+        number_of_reads = s.get(ELEMENT_NB_BASE_R1) + s.get(ELEMENT_NB_BASE_R2)
+        return (reads_q30/number_of_reads) * 100
```
tcezard: These lines should go but FYI these values are number of bases not number of reads

analysis_driver/dataset.py
```diff
@@ -498,10 +498,10 @@ def _pipeline_instruction(self):
         return instruction
 
     def _is_ready(self):
-        return self.data_threshold and int(self._amount_data()) > int(self.data_threshold)
+        return self.data_threshold and self.pc_q30 > 75 and int(self._amount_data()) > int(self.data_threshold)
 
     def report(self):
-        runs = sorted(set(r.get(ELEMENT_RUN_NAME) for r in self.run_elements))
+        runs = self.sample.get('run_ids')
```
tcezard: `run_id` is not part of  `sample`. It should be `sample.get('aggregated').get('run_ids')`

tests/test_dataset.py
```diff
@@ -402,9 +392,26 @@ def setup_dataset(self):
...
+        ]
+
+        self.dataset._sample = {
+                                'aggregated': {'clean_yield_q30_in_gb': 1.5},
+                                'required_yield': 1000000000,
+                                'q30_bases_r1': 60000,
+                                'q30_bases_r2': 50000,
+                                'bases_r1': 70000,
+                                'bases_r2': 70000,
+                                'run_ids': ['a_run_id', 'another_run_id']
```
tcezard: `run_id` is in `aggregated`

tcezard (changes requested):

analysis_driver/dataset.py
```diff
@@ -468,22 +466,21 @@ def non_useable_run_elements(self):
...
+        return self.sample['project_id']
 
     def _amount_data(self):
-        return sum(
-            [
-                int(r.get(ELEMENT_NB_Q30_R1_CLEANED, 0)) + int(r.get(ELEMENT_NB_Q30_R2_CLEANED, 0))
-                for r in self.run_elements
-            ]
-        )
+        return int(self.sample.get('aggregated').get('clean_yield_in_gb')*1000000000)
```
tcezard, line 481: The casting to `int` here is pretty useless. Check if `self.sample.get('aggregated').get('clean_yield_in_gb')` returns an `int` and if it does not then cast it

analysis_driver/dataset.py
```diff
@@ -498,10 +495,10 @@ def _pipeline_instruction(self):
         return instruction
 
     def _is_ready(self):
-        return self.data_threshold and int(self._amount_data()) > int(self.data_threshold)
+        return self.data_threshold and self.pc_q30 > 75 and int(self._amount_data()) > int(self.data_threshold)
```
tcezard: The casting is probably not necessary here either. Check with real data.

analysis_driver/dataset.py
```diff
@@ -513,6 +510,7 @@ def report(self):
             s += '(no non useable run elements)'
         return s
 
+
```
tcezard: unnecessary new line

tcezard (changes requested):
few more changes, but nothing major

analysis_driver/dataset.py
katieemelianova: When I checked this it returned a float, which is why it is cast to an int here

tcezard: That makes sense

tcezard (approved):

PR #312, opened 2018-01-22
Estimate memory GenotypeGVCFs will use

katieemelianova:
Change memory requirements in GenotypeGVCFs to be either 50 (minimum) or 3 times the number of samples


analysis_driver/quality_control/relatedness.py
```diff
@@ -154,10 +154,14 @@ class GenotypeGVCFs(RelatednessStage):
     gVCFs = ListParameter()
     reference = Parameter()
 
+    @property
+    def memory(self):
+        return max(len(self.gVCFs) * 3, 50)
+
     def gatk_genotype_gvcfs_cmd(self):
```
tcezard: I would prefer if `number_threads` and `memory` were parameters of `gatk_genotype_gvcfs_cmd` and passed from `_run` function

analysis_driver/quality_control/relatedness.py
```diff
@@ -168,7 +172,7 @@ def _run(self):
             job_name='gatk_genotype_gvcfs',
             working_dir=self.job_dir,
             cpus=12,
-            mem=96
+            mem=self.memory
```
tcezard: make sure the memory passed to slurm is slightly greater than the memory passed to java (+2G maybe)

tcezard (changes requested):

tcezard (approved):
That looks good

PR #316, opened 2018-01-31
Fix for sample's default values scanner forcing data threshold

### LINKED ISSUE #315: Data threshold is set in the DatasetScanner
In SampleDataset, the data thresholds for status ready has changed to used required yield.
However it was also set in the dataset scanner which means that it is still uses the required yield Q30

tcezard:
This PR sets default values for `pc_q30` and `_amount_data` in `SampleDataset` which superseded #314 
It also fixes #315 by removing the query for data threshold in SampleScanner.

### DIFF
--- a/analysis_driver/dataset_scanner.py
+++ b/analysis_driver/dataset_scanner.py
@@ -48,18 +48,18 @@ class DatasetScanner(AppLogger):
             statuses.remove(DATASET_NEW)
             statuses.append(None)
         if len(statuses) > 1:
-            match = {'$or': [{'proc_status': status} for status in statuses]}
+            where = {'$or': [{'aggregated.most_recent_proc.status': status} for status in statuses]}
         else:
-            match = {'proc_status': statuses[0]}
+            where = {'aggregated.most_recent_proc.status': statuses[0]}
         return [
-            d for d in get_documents(self.endpoint, match=match, paginate=False, quiet=True)
+            d for d in get_documents(self.endpoint, where=where, all_pages=True, quiet=True, max_results=100)
             if d[self.item_id] not in self._triggerignore
         ]
 
     def _get_datasets_for_statuses(self, statuses):
         self.debug('Creating Datasets for status %s', ', '.join(statuses))
         return [
-            self.get_dataset(d[self.item_id], d.get('most_recent_proc'))
+            self.get_dataset(d[self.item_id], d.get('aggregated', {}).get('most_recent_proc'))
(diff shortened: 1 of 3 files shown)


katieemelianova (approved):

analysis_driver/dataset_scanner.py
```diff
@@ -48,18 +48,18 @@ def _get_dataset_records_for_statuses(self, statuses):
...
             statuses.append(None)
         if len(statuses) > 1:
-            match = {'$or': [{'proc_status': status} for status in statuses]}
+            where = {'$or': [{'aggregated.most_recent_proc.status': status} for status in statuses]}
         else:
-            match = {'proc_status': statuses[0]}
+            where = {'aggregated.most_recent_proc.status': statuses[0]}
         return [
-            d for d in get_documents(self.endpoint, match=match, paginate=False, quiet=True)
+            d for d in get_documents(self.endpoint, where=where, all_pages=True, quiet=True, max_results=100)
```
mwhamgenomics, line 55: Will we still be able to report on sample repeats with max_results=100?

analysis_driver/dataset_scanner.py
```diff
@@ -129,33 +129,15 @@ def _is_valid_dataset(self, dataset):
...
-                req_yield = sample.udf.get('Yield for Quoted Coverage (Gb)') * 1000000000
-                datasets[sanitize_user_id(sample.name)]['threshold'] = req_yield
-            self.debug('Lims query complete')
-
-        return [
-            self.get_dataset(k, v['record'].get('most_recent_proc'), v.get('threshold'))
-            for k, v in datasets.items()
-        ]
+    def get_dataset(self, name, most_recent_proc=None):
+        return SampleDataset(name, most_recent_proc)
```
mwhamgenomics, line 136: This removes previous optimisation, so a new query will be made for every sample scanned. We should re-add the optimisation later

analysis_driver/dataset_scanner.py
tcezard: Yes rest_communication will get all pages with option `all_pages=True`

analysis_driver/dataset_scanner.py
tcezard: That is correct. However because we're not using aggregate endpoint anymore it is still a faster.
I agree that the optimisation should be added  back in next version and probably extended to run and projects.

mwhamgenomics (approved):

tcezard:
The build is failing due to dependency issues but is passing otherwise.

PR #318, opened 2018-02-05
Fix pipeline resume

### LINKED ISSUE #293: Resume option does not work
I've tried to us the `--resume` option on analysis driver and it does not work at the moment.
 - Running `--resume` does not change the `analysis_driver_procs`'s status to resume.
 - After manually changing the  `analysis_driver_procs`'s status to resume manually it is not picked up and restarted by the pipeline.

mwhamgenomics:
Fixes pipeline resuming by adding `PIPELINE_RESUME` to client, dataset_scanner, etc, doing a retrieve_entity in `dataset.start`, and setting `self.proc_id` in `MostRecentProc.retrieve_entity`. Adding an integration test for it.

Closes #293


analysis_driver/client.py
```diff
@@ -61,8 +61,12 @@ def main(argv=None):
             scanner.report(all_datasets=True)
         return 0
 
-    datasets = scanner.scan_datasets(DATASET_NEW, DATASET_REPROCESS, DATASET_READY, DATASET_FORCE_READY)
-    ready_datasets = datasets.get(DATASET_FORCE_READY, []) + datasets.get(DATASET_READY, [])
+    processable_statuses = (DATASET_FORCE_READY, DATASET_READY, DATASET_RESUME)
```
tcezard: Can we swap `DATASET_READY` and `DATASET_RESUME` so that the resumed dataset take priority over the ready.

integration_tests/integration_test.py
```diff
@@ -84,7 +85,7 @@ def setUp(self):
         rest_communication.post_entry(
             'samples',
             {'library_id': self.library_id, 'project_id': self.project_id, 'sample_id': self.sample_id,
-             'run_elements': [e['run_element_id'] for e in run_elements], 'required_yield': 100000000}
+             'run_elements': [e['run_element_id'] for e in run_elements], 'required_yield_q30': 900000000}
```
tcezard: Not sure why we're setting the `required_yield_q30` here instead of the `required_yield`

tcezard (changes requested):

tcezard (approved):

PR #319, opened 2018-02-06
Nonhuman variantcalling

katieemelianova:
Add variant calling to QC pipeline, and extract ti/tv and het/hom ratio from results
Also extract ti/tv and het/hom ratio from result of variant_calling pipeline
report these metrics in all sample pipelines


analysis_driver/pipelines/variant_calling.py
```diff
@@ -129,20 +171,26 @@ def _run(self):
...
             ext=('--pair_hmm_implementation VECTOR_LOGLESS_CACHING -ploidy 2 --emitRefConfidence GVCF '
-                 '--variant_index_type LINEAR --variant_index_parameter 128000 --dbsnp ' + self.dbsnp)
+                 '--variant_index_type LINEAR --variant_index_parameter 128000 ')
         )
         for annot in ('BaseQualityRankSumTest', 'FisherStrand', 'GCContent', 'HaplotypeScore',
                       'HomopolymerRun', 'MappingQualityRankSumTest', 'MappingQualityZero', 'QualByDepth',
                       'ReadPosRankSumTest', 'RMSMappingQuality', 'DepthPerAlleleBySample', 'Coverage',
                       'ClippingRankSumTest', 'DepthPerSampleHC'):
             haplotype_cmd += ' --annotation ' + annot
+        if self.dbsnp:
```
mwhamgenomics, line 170: If we don't have dbsnp for the species or if we forget to add it to the config, we'll be missing out on some annotation without knowing. Are we sure about this?

analysis_driver/pipelines/variant_calling.py
```diff
@@ -36,6 +42,29 @@ def gatk_cmd(self, run_cls, input_bam, output, xmx=16, nct=16, ext=None):
             output=output
         )
 
+    @property
+    def select_snp_command(self):
```
mwhamgenomics: This can be pushed down to SelectVariants

analysis_driver/pipelines/variant_calling.py
```diff
@@ -36,6 +42,29 @@ def gatk_cmd(self, run_cls, input_bam, output, xmx=16, nct=16, ext=None):
...
+        command = self.gatk_cmd('SelectVariants', self.raw_snp_vcf, nct=1, nt=16)
+        command += ' -V ' + self.genotyped_vcf
+        command += ' -selectType SNP '
+        return command
+
+    @property
+    def filter_snp_command(self):
```
mwhamgenomics: This can be pushed down to VariantFiltration

analysis_driver/pipelines/variant_calling.py
```diff
@@ -152,10 +200,43 @@ def _run(self):
...
+            working_dir=self.gatk_run_dir,
+            mem=16
+        ).join()
+
+
+class SelectVariants(GATKStage):
+    command = Parameter()
+    def _run(self):
+        return executor.execute(
+            self.command,
```
mwhamgenomics: This doesn't need to be a parameter - can just use self.select_snp_command

analysis_driver/pipelines/variant_calling.py
```diff
@@ -152,10 +200,43 @@ def _run(self):
...
+            working_dir=self.gatk_run_dir,
+            mem=16
+        ).join()
+
+
+class VariantFiltration(GATKStage):
+    command = Parameter()
+    def _run(self):
+        return executor.execute(
+            self.command,
```
mwhamgenomics: This doesn't need to be a parameter - can just use self.filter_snp_command

analysis_driver/pipelines/variant_calling.py
```diff
@@ -179,14 +260,18 @@ def build_pipeline(dataset):
     def stage(cls, **params):
         return cls(dataset=dataset, **params)
 
+    g = GATKStage(dataset=dataset)
```
mwhamgenomics: We shouldn't really be creating instances of abstract classes like this. With the fixes above, it should be possible to remove this

analysis_driver/pipelines/qc.py
```diff
@@ -1,12 +1,25 @@
...
 toolset_type = 'non_human_sample_processing'
 name = 'qc'
 
 
 def build_pipeline(dataset):
+
+    def stage(cls, **params):
+        return cls(dataset=dataset, **params)
+
+    g = v.GATKStage(dataset=dataset)
```
mwhamgenomics, line 13: Kind of annoying that this is necessary, but works for now. Should probably look at cleaning up the handling of pipeline files later.

analysis_driver/pipelines/qc.py
```diff
@@ -1,12 +1,25 @@
...
     bam_file_production = common.build_bam_file_production(dataset)
-    data_output = common.SampleDataOutput(dataset=dataset, previous_stages=bam_file_production, output_fileset='non_human_qc')
+    haplotype = stage(v.HaplotypeCaller, input_bam=g.sorted_bam, previous_stages=bam_file_production)
+    genotype = stage(v.GenotypeGVCFs, previous_stages=[haplotype])
+    select_snp = stage(v.SelectVariants, command=g.select_snp_command, previous_stages=[genotype])
+    filter_snp = stage(v.VariantFiltration, command=g.filter_snp_command, previous_stages=[select_snp])
+    vcfstats = stage(qc.VCFStats, vcf_file=g.filter_snp_vcf, previous_stages=[filter_snp])
+    bgzip = stage(v.BGZip, previous_stages=[vcfstats])
+    tabix = stage(v.Tabix, previous_stages=[bgzip])
+    data_output = common.SampleDataOutput(dataset=dataset, previous_stages=[tabix], output_fileset='non_human_qc')
```
mwhamgenomics: This can be a call to `stage()`

analysis_driver/pipelines/variant_calling.py
```diff
@@ -152,10 +200,43 @@ def _run(self):
         ).join()
 
 
+class GenotypeGVCFs(GATKStage):
```
mwhamgenomics, line 183: This basically duplicates relatedness.GenotypeGVCFs - should refactor later

mwhamgenomics (changes requested):

tcezard (changes requested):
Can you add unit tests at least for the new Classes/functions/properties.
It would also be goo to take the opportunity to add tests for other Stages

analysis_driver/pipelines/variant_calling.py
```diff
@@ -63,19 +70,48 @@ def indel_realigned_bam(self):
...
+        if not dbsnp and self.dataset.pipeline.name == 'variant_calling':
+            raise AnalysisDriverError('Could not find dbsnp file for %s' % self.dataset.name)
+        return dbsnp
 
     @property
     def known_indels(self):
         return cfg.query('genomes', self.dataset.genome_version, 'known_indels')
 
 
+
```
mwhamgenomics: ?

analysis_driver/pipelines/variant_calling.py
```diff
@@ -63,19 +70,48 @@ def indel_realigned_bam(self):
...
+
+
+
+
+
+
+
+
+
```
mwhamgenomics: ?

mwhamgenomics (approved):

tcezard (approved):

PR #326, opened 2018-03-26
Add source of constituent data to projects and samples

katieemelianova:
Add data source to samples (run elements included) and projects (samples included)

Add tests


analysis_driver/dataset.py
```diff
@@ -657,6 +669,10 @@ def initialise_entity(self):
             'dataset_type': self.dataset.type,
             'dataset_name': self.dataset.name
         }
+
+        if self.dataset.type in ['project', 'sample']:
+            entity.update({'data_source': self.dataset.data_source})
```
tcezard: This is going to raise NotImplementedError in RunDataset's case

tcezard (changes requested):
`data_source` needs to be implemented in RunDataset

analysis_driver/dataset.py
katieemelianova: This I think would not raise NotImplementedError in RunDataset due to the if statement - unless I'm missing something?
I have added data_source to RunDataset but returning None instead to not have to check for dataset type

mwhamgenomics (approved):

tcezard (approved):

analysis_driver/dataset.py
tcezard: that is indeed correct

PR #331, opened 2018-05-30
Remove phix

### LINKED ISSUE #329: Remove Phix from raw data
The RunProcessing pipeline need to be able to remove reads identified as Phix and remove them from the raw data.
We will align the reads to the phix genome to identify and remove them from the raw data after bcl2fastq using [FastqFilterer](https://github.com/EdinburghGenomics/Fastq-Filterer) after [this issue](https://github.com/EdinburghGenomics/Fastq-Filterer/issues/12) has been addressed.

CR840

### LINKED ISSUE #330: Remove Phix reads from processed data
Some fastq files still contains Phix reads.
We need to remove the Phix reads by identifying them and filtering the fastq file.
then we need to rerun the run processing QC and update the run metrics on the reporting app 
CR840

tcezard:
This PR will add a new step in demultiplexing pipeline that will align read1 to the Phix genome. get the reads that align and extract them from the fastq file using fastqfilterer's new feature.
fixes #329 
It also add a new script that can take already generated fastq file and extract PhiX reads using the same procedure. The script then rerun the QCs steps and upload the results to the reporting app.
fixes #330 
CR840

### DIFF
--- a/analysis_driver/pipelines/demultiplexing.py
+++ b/analysis_driver/pipelines/demultiplexing.py
@@ -94,12 +94,29 @@ class Bcl2Fastq(DemultiplexingStage):
         return bcl2fastq_exit_status
 
 
-class FastqFilter(DemultiplexingStage):
-    def _run(self):
+class PhixDetection(DemultiplexingStage):
 
+    def _run(self):
         # Send the results of BCL2fastq to the rest API
         crawler = RunCrawler(self.dataset, run_dir=self.fastq_dir, stage=RunCrawler.STAGE_CONVERSION)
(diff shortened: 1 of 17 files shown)


mwhamgenomics (approved):
There's an assumption that runs will always have PhiX, which is not always the case. Everything should still work, there will just be empty read name files - can clean that up at a later date.

Kind of annoying that we're basically duplicating half the run processing pipeline in `remove_phix.py`, especially since Luigi supports resolving task dependency trees up to certain endpoints. We should have a look at addressing that later.

PR #337, opened 2018-06-11
Fix gatk genotype

### LINKED ISSUE #336: genotype gvcf takes too much memory for genomes with many contigs
Solution include:
 - reducing the number of contigs in genomes
 - increasing the amount of memory available
 - split the genome in pieces when processing 

It might be that changing the indexing from GATK builting to tabix might help as well. 
EDIT:
It also seems that lustre prevent the use of the GATK builtin index. using tabix index seems then required

tcezard:
fixes #336


tcezard:
@mwhamgenomics This is ready for review.
Integration test are fine.

mwhamgenomics (approved):

Issue #249, opened 2017-05-16
Split Dataset and Pipeline run data


mwhamgenomics:
The Dataset is quite big now, and is serving two purposes - managing data (species, run elements, etc.), and managing the pipeline run (notifications, exceptions, etc.). We should split the pipeline functionality off into a new class, e.g. `PipelineRun`.

PR #338, opened 2018-06-13
Eager loading the output config file

### LINKED ISSUE #282: output_files.yaml is lazily loaded into Analysis driver code
output_files.yaml is loaded when `OutputFileConfiguration` is created which can happen at the end of the process. This can result in discrepancy between the code/config loaded at the beginning of the process and the output_files.yaml loaded at the end in cases where a new version of analysis driver is deployed.
We could use a singleton of OutputFileConfiguration to fix this.

tcezard:
Ensure the output config is loaded early on in the pipeline. 
This also limit the config to only one type of pipeline output.
Fixes #282


analysis_driver/pipelines/common.py
```diff
@@ -85,7 +85,8 @@ class SampleDataOutput(segmentation.Stage):
 
     @property
     def output_cfg(self):
-        return OutputFileConfiguration(self.output_fileset)
+        output_file_config.set_pipeline_type(self.output_fileset)
+        return output_file_config
```
mwhamgenomics, line 89: Having this as a property probably isn't the cleanest solution anymore - can fix in #249

mwhamgenomics (approved):

PR #347, opened 2018-09-14
Production fixes, integration testing

### LINKED ISSUE #291: Repeated stage running
Sometimes we want to be able to run stages, e.g. crawlers, multiple times in the pipeline, which at the moment requires the definition of a new class. This can be made easier by making `__stagename__` a Parameter. We should also add something to the pipeline build tests to ensure no two stages have the same stagename.

### LINKED ISSUE #292: More verbose comparisons in integration tests
At the moment in integration tests, we have no record of the observed/expected data in successful comparisons. We should either log the contents of the config or verbosely log each comparison.

### LINKED ISSUE #311: Remove all usage of aggregate endpoints
the aggregate endpoint will be deprecated in future release of Reporting app.
We should switch all usage to the aggregated sub field in the Eve endpoints

### LINKED ISSUE #334: Analysis Driver Process doesn't record start_date
Reporting app schema defines a `start_date` for `analysis_driver_proc` but it is not populated by Analysis Driver.
The start date should be set on the pipeline start but not on resume.

### LINKED ISSUE #342: Memory allocation for indel realignment
Memory should be increased 32 GB.

### LINKED ISSUE #343: Deterministic file checksums in integration tests
#292 will introduce integration testing via EGCG-Core, which runs the tests inside a uniquely-named temp folder. This means that any output files containing file paths will be non-deterministic, meaning that it will be impossible to check for expected MD5s. During MD5 checks, we could possibly:

- read the file in chunks
- in each chunk, remove any instances of the temp file path (i.e. the working directory), or replace it with a placeholder
- pipe the result into MD5

mwhamgenomics:
- Making failure notifications more fault-tolerant
- Removing aggregate queries - closes #311 
- Adding date_started to MostRecentProc.start, except on resume - closes #334 
- BCLValidator construction fix in demultiplexing
- Making Stage.stage_name a parameter - closes #291 
- Applying production fixes
  - indel realignment memory - closes #342 
  - relatedness memory limit
  - fix_phix error handling
- Adding integration testing with EGCG-Core - closes #292, closes #343


analysis_driver/dataset.py
```diff
@@ -316,27 +316,25 @@ def run_elements(self):
             self._run_elements = self._run_elements_from_lims()
         return self._run_elements
 
+    @staticmethod
+    def _find_pooling_step_for_artifact(art, max_iterations=10):
```
tcezard: What is  the rational for removing the `expected_pooling_step_name`?

analysis_driver/pipelines/demultiplexing.py
```diff
@@ -28,14 +28,13 @@ def _run(self):
...
-        if not isdir(self.fastq_dir):
-            makedirs(self.fastq_dir)
+        makedirs(self.fastq_dir, exist_ok=True)
 
         # Send the information about the run to the rest API
         crawler = RunCrawler(self.dataset)
         crawler.send_data()
 
-        b = BCLValidator(self.job_dir, self.dataset)
+        b = BCLValidator(dataset=self.dataset)
```
tcezard, line 37: How did this even work before?

integration_tests/integration_test.py
```diff
@@ -1,84 +1,62 @@
...
         for lane in range(1, 9):
             run_elements.append(
                 {'run_id': self.run_id, 'project_id': self.project_id, 'sample_id': self.sample_id,
                  'library_id': self.library_id, 'run_element_id': '%s_%s_%s' % (self.run_id, lane, self.barcode),
-                 'useable': 'yes', 'barcode': self.barcode, 'lane': lane, 'clean_q30_bases_r1': 57000000,
-                 'clean_q30_bases_r2': 57000000, 'clean_reads': 1, 'clean_bases_r1': 60000000, 'clean_bases_r2': 60000000}
+                 'useable': 'yes', 'barcode': self.barcode, 'lane': lane, 'bases_r1': 60000000, 'bases_r2': 60000000,
+                 'clean_bases_r1': 58000000, 'clean_bases_r2': 58000000, 'q30_bases_r1': 57000000,
+                 'q30_bases_r2': 57000000, 'clean_q30_bases_r1': 56000000, 'clean_q30_bases_r2': 56000000,
+                 'clean_reads': 1}
```
tcezard, line 59: Why do you need to change the run_elements values ?

integration_tests/integration_test.py
```diff
@@ -93,25 +71,38 @@ def setUp(self):
...
-        os.mkdir(cfg['jobs_dir'])
-        os.mkdir(cfg[test_type]['output_dir'])
+        patches = (
+            patch('egcg_core.clarity.get_sample', new=_fake_get_sample),
+            patch('egcg_core.clarity.get_species_from_sample', return_value=species)
+        )
+        for p in patches:
+            p.start()
+
+        self.patches += patches
```
tcezard: Is this not going to accumulate patches over multiple tests or are they somehow getting removed somewhere?

analysis_driver/pipelines/demultiplexing.py
mwhamgenomics: This worked because by coincidence, the Luigi parameters were getting correctly assigned as positional arguments to `self.previous_stages` (which we weren't using because we weren't using BCLValidator.run) and `self.dataset`

integration_tests/integration_test.py
mwhamgenomics: We had to add `q30_bases` for the new aggregation, but it's not important what exact values these are - this is input data in order for the sample pipeline to pick up enough yield that it processes the sample

integration_tests/integration_test.py
mwhamgenomics: This will not happen because a new IntegrationTest object is constructed per test method run. Can move these extra patches to their own attribute to make it more explicit

analysis_driver/dataset.py
mwhamgenomics: This was because the argument was always 'Create PDP Pool'. Now that this is no longer a nested function, it can be more general-purpose - will reinstate the argument

tcezard (approved):

PR #351, opened 2018-09-27
Overwrite with bgzip and tabix

### LINKED ISSUE #340: force overwrite in bgzip and Tabix steps
At the moment bgzip and tabix are set to not overwrite their respective output file.
We should change this behaviour to always overwrite.

tcezard:
make all tabix/bgzip command overwrite their output file if they exist already
Also refactor the commands to all use the same bash command generation.
fixes #340


mwhamgenomics (approved):

tcezard:
Validation ID: VB1

PR #353, opened 2018-10-04
Project on valid project

### LINKED ISSUE #341: Trio checking tries to process incompatible projects
At the moment, the project pipeline runs completed projects indiscriminately, including non-human projects with only fastqs generated. We should fix this.

### LINKED ISSUE #352: Run crawler crashes if the insert size metric is missing
If the insert size in orientation FR is missing the RunCrawler will crash with KeyError.
https://github.com/EdinburghGenomics/Analysis-Driver/blob/b635c498126bf374f2774438db3c1ea04663176b/analysis_driver/report_generation/run_crawler.py#L342
Adding an if allows it to upload the information available.

tcezard:
This PR makes the project process run an empty pipeline on projects that do not have enough gvcf.
Check if a sample should have a gvcf based on the pipeline it was processed with's name 

Also add in project process integration tests
closes #341 
Also add bugfix for cases when insert size metric is missing: close #352

### DIFF
--- a/analysis_driver/dataset.py
+++ b/analysis_driver/dataset.py
@@ -708,7 +708,7 @@ class MostRecentProc:
                 }
             }
             if self.get('status') != DATASET_RESUME:
-                payload['date_started'] = now()
+                payload['start_date'] = now()
 
             self.update_entity(status=DATASET_PROCESSING, pid=os.getpid())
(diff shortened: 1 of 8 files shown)


tcezard:
Validation id: VB1

mwhamgenomics (approved):

PR #354, opened 2018-10-11
Faster QC using only a subset of the chromosomes

### LINKED ISSUE #344: Replace GATK base variant calling with faster alternative in QC
The variant call QC steps for poorly annotated genome is taking a long time to finish.
Alternative tools are samtools/freebayes which would be much faster.

tcezard:
Java commands now set the time using local timezone
Add optional intervals to GATK commands
Use precomputed interval of a maximum 100Mbases for Haplotype caller QC
fixes #344


tcezard:
This implemetation won't work. Closing for now

PR #360, opened 2018-11-22
Update EGCG-Core to 0.9.1

### LINKED ISSUE #359: SSL error when when multiple stages start at the same time
This is an issue with rest_communication
see https://github.com/EdinburghGenomics/EGCG-Core/issues/91

tcezard:
fixes #359

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-EGCG-Core==0.9
+EGCG-Core==0.9.1
 luigi==2.5.0
 ete3==3.0.0b35
 pandas<0.21


tcezard:
This new version of EGCG-Core brings a fix to rest communication when used in the context of multiprocessing with https connections.
The bugs prevents some concurrent pipeline stages from starting at the same time which is high impact. The fixes does not change any functionality beyond the bug fix and is, as a result, low risk.
This changes will by-pass validation

mwhamgenomics (approved):

PR #362, opened 2018-11-28
Support for multiprocessing when raising exceptions

### LINKED ISSUE #335: Aborted sequencing runs are failing
They should be marked as aborted and have the cleanup and Automatic run review performed.

tcezard:
Exception raised in stages are now stored in a multiprocessing.Queue.
In order to receive them when multiprocessing is active
This includes PipelineError which is raised when a stage returns non 0 exist code.
This will closes #335 but is not very elegant, especially the fact that there are Run specific action in the `client.py` which should be process type agnostic.
If you can think of a better solution do let me know.

### DIFF
--- a/analysis_driver/dataset.py
+++ b/analysis_driver/dataset.py
@@ -1,3 +1,4 @@
+import multiprocessing
 import os
 import re
 import signal
@@ -6,7 +7,6 @@ from datetime import datetime
 from errno import ESRCH
 from sys import modules
 from time import sleep
-from collections import OrderedDict
 from egcg_core import rest_communication, clarity
 from egcg_core.app_logging import AppLogger
 from egcg_core.config import cfg
@@ -28,7 +28,7 @@ class Dataset(AppLogger):
     type = None
     endpoint = None
     id_field = None
-    exceptions = OrderedDict()
+    exceptions = multiprocessing.Queue()  # This needs to be a Queue to be accessible from every subprocess
 
     def __init__(self, name, most_recent_proc=None):
         self.name = name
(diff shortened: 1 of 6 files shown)


analysis_driver/dataset.py
```diff
@@ -160,21 +160,23 @@ def _is_ready(self):
...
     def register_exception(self, luigi_task, exception):
-        self.exceptions[luigi_task.stage_name] = exception
+        self.exceptions.put((luigi_task.stage_name, exception))
 
     def raise_exceptions(self):
-        if self.exceptions:
-            self.critical('%s exceptions registered with dataset %s', len(self.exceptions), self.name)
-            for name in self.exceptions:
+        if not self.exceptions.empty():
+            first_exception = None
```
lukbut, line 167: What are these four lines of code doing?

tests/test_pipelines/test_pipeline.py
```diff
@@ -0,0 +1,55 @@
...
+                p.stop()
+
+    def test_pipeline_with_exceptions(self):
+        with pytest.raises(SequencingRunError):
+            self._pipeline_with_stage(ExceptionStage)
+
+    def test_successful_pipeline(self):
+        assert self._pipeline_with_stage(SuccessfulStage) == 0
+
+    def test_failling_pipeline(self):
```
lukbut:
```suggestion
    def test_failing_pipeline(self):
```

lukbut: Just a little typo in "failing"

tests/test_pipelines/test_pipeline.py
```diff
@@ -0,0 +1,55 @@
...
+    def _run(self):
+        return 1
+
+
+class TestPipeline(TestAnalysisDriver):
+
+    def _pipeline_with_stage(self, stage_class):
+        dataset = RunDataset(name='a_sample')
+        patches = []
+        for function_to_patch in ['start', 'get_stage', 'start_stage', 'end_stage', 'resolve_pipeline_and_toolset']:
```
lukbut, line 32: What is happening here?

lukbut (changes requested):

analysis_driver/dataset.py
tcezard: This loop will iterate over all the registered exceptions and log them all.
It also keep the first on to raise it at the end.
We only expect one exception to be raise in the children process.

tests/test_pipelines/test_pipeline.py
tcezard: We're patching all the functions from `RunDataset` that will contact the REST API.

lukbut (approved):

integration_tests/integration_test.py
```diff
@@ -226,14 +233,19 @@ def test_demultiplexing(self):
...
-            self.expect_equal(
-                rest_communication.get_document('analysis_driver_procs')['status'], 'aborted'
-            )
+        self.assertEqual('exit status', exit_status, 0)
+        ad_proc = rest_communication.get_document('analysis_driver_procs', where={'dataset_name': self.run_id})
+
+        self.expect_equal(
+            ad_proc['status'], 'aborted', 'pipeline status'
+        )
+        self.expect_stage_data([('setup', 9)], analysis_driver_proc=ad_proc['proc_id'])
```
mwhamgenomics: This keyword arg won't work - should be `where={'analysis_driver_proc':...`

mwhamgenomics (changes requested):

integration_tests/integration_test.py
tcezard: This is fixed now. Tests did pass cause there is only one stage at this point

tcezard:
Validation ID: VB3

mwhamgenomics (approved):

PR #364, opened 2018-12-06
Embedded procs, integration test fixes

### LINKED ISSUE #284: Missing analysis_driver_procs from runs
In dataset.py, `initialise_entity` is trying to patch the run with the new analysis_driver_proc at a time that the run does not yet exist. The run is created by the RunCrawler at the end of the pipeline, which doesn't help much when the dataset is trying to patch it at the start. This means we're missing embedded pipeline processes. We should do something to initialise container entities (runs, samples, projects) at the start of the pipeline.

mwhamgenomics:
- Closes #284
- Fixes integration tests

### DIFF
--- a/analysis_driver/report_generation/run_crawler.py
+++ b/analysis_driver/report_generation/run_crawler.py
@@ -3,7 +3,7 @@ from os.path import isfile
 from collections import defaultdict, Counter
 from egcg_core import util
 from egcg_core.constants import *
-from egcg_core.rest_communication import post_or_patch as pp
+from egcg_core.rest_communication import patch_entry, post_or_patch as pp
 from analysis_driver.reader import demultiplexing_parsers as dm, mapping_stats_parsers as mp
 from analysis_driver.exceptions import PipelineError
 from .crawler import Crawler
@@ -51,7 +51,7 @@ class RunCrawler(Crawler):
         self.unexpected_barcodes = {}
         self.libraries = defaultdict(dict)
         self.lanes = defaultdict(dict)
-        self.run = {ELEMENT_RUN_NAME: self.dataset.name, ELEMENT_RUN_ELEMENTS: []}
+        self.run = {ELEMENT_RUN_ELEMENTS: []}
         self.projects = defaultdict(dict)
 
         for run_element in dataset.run_elements:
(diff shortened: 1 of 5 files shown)


analysis_driver/report_generation/run_crawler.py
```diff
@@ -3,7 +3,7 @@
 from collections import defaultdict, Counter
 from egcg_core import util
 from egcg_core.constants import *
-from egcg_core.rest_communication import post_or_patch as pp
+from egcg_core.rest_communication import patch_entry, post_or_patch as pp
```
lukbut, line 6: @tcezard @mwhamgenomics Should we be using patch? I have never used it in the past personally, and I've ran a quick search and have found the IETF guideline which states that patch is not safe to use. 
https://tools.ietf.org/html/rfc5789

tcezard: The word safe in this document just means that the method can change existing data which is the point of PATCH and PUT.
See https://stackoverflow.com/questions/41390997/why-patch-is-neither-safe-nor-idempotent
for a better explanation. Also note that Eve uses etags to ensure you're updating the latest version of the resource.

tcezard (approved):

analysis_driver/report_generation/run_crawler.py
lukbut: @tcezard What do you mean by Eve uses etags, not sure I understand?

tcezard: See http://docs.python-eve.org/en/latest/features.html?highlight=etag#data-integrity-and-concurrency-control

lukbut: Thanks @tcezard

tcezard:
Validation ID: VB3

mwhamgenomics:
Integration tests passing on mwham_refactor, which was created from this branch:
```
------
Integration test finished
Run on commit f2fa316 (HEAD, origin/mwham_refactor, mwham_refactor), made on Wed, 12 Dec 2018 15:36:15 +0000

Start time: 2018-12-12_15:39:00, finish time: 2018-12-13_14:08:52
Pytest output:
================= test session starts ==================
platform linux -- Python 3.4.4, pytest-3.6.0, py-1.5.3, pluggy-0.6.0
rootdir: /lustre/home/mwham/integration_test_fE8bE0/Analysis-Driver, inifile:
plugins: xdist-1.22.2, forked-0.2, cov-2.5.1
collected 6 items

integration_tests/integration_test.py ......                             [100%]

=============== 6 passed in 80991.70 seconds ===============
```

PR #368, opened 2018-12-18
Early alignment in Run processing 

### LINKED ISSUE #357: Early alignment during run processing
Alignment during run processing could be done before the end of the run which would allows the demultiplexing to be performed faster.
This will involve 
- early demultiplexing step with 50 bases of R1 and 50 bases of R2
- alignment of the reads as is currently
- insert size and duplicate calculation

Validation should look at the QC obtained previously and ensure results are similar in value

tcezard:
This PR move the alignments of fastq file before the run finishes.
It allows the alignment to start and potentially finish before the run is completed.
This should save a significant amount of time.
closes #357


analysis_driver/pipelines/demultiplexing.py
```diff
@@ -223,7 +226,50 @@ def _run(self):
...
+        current_cycle = sorted(interop_metrics.get_cycles_extracted(self.dataset.input_dir))[-1]
+        while current_cycle < required_number_of_cycle and self.dataset.is_sequencing():
+            time.sleep(1200)
+            current_cycle = sorted(interop_metrics.get_cycles_extracted(self.dataset.input_dir))[-1]
+        return 0
+
+
+class HalfRun(DemultiplexingStage):
+    @property
+    def fastq_intermetiate_dir(self):
```
mwhamgenomics: intermetiate -> intermediate

analysis_driver/pipelines/demultiplexing.py
```diff
@@ -223,7 +226,50 @@ def _run(self):
...
+
+        # get the last cycle extracted
+        current_cycle = sorted(interop_metrics.get_cycles_extracted(self.dataset.input_dir))[-1]
+        while current_cycle < required_number_of_cycle and self.dataset.is_sequencing():
+            time.sleep(1200)
+            current_cycle = sorted(interop_metrics.get_cycles_extracted(self.dataset.input_dir))[-1]
+        return 0
+
+
+class HalfRun(DemultiplexingStage):
```
mwhamgenomics: We're talking about half runs here, but we take 50 bases, i.e. a third of the run. Maybe use a more generic term here, e.g. `PartialRun`

analysis_driver/pipelines/demultiplexing.py
```diff
@@ -251,6 +297,14 @@ def fastq_base(self, run_element):
         fastq_pair = self.fastq_pair(run_element)
         return fastq_pair[0][:-len('_R1_001.fastq.gz')]
 
+    def real_fastq_base(self, run_element):
```
mwhamgenomics: This name is a bit confusing next to `fastq_base` - what's the difference between them?

mwhamgenomics (changes requested):

analysis_driver/pipelines/demultiplexing.py
tcezard: fixed

analysis_driver/pipelines/demultiplexing.py
tcezard: Changed

analysis_driver/pipelines/demultiplexing.py
tcezard: Names have been updated to be more descriptive:
 - `intermediate_fastq_base` refers to the location of the fastq for the partial run
 - `final_fastq_base` refers to the location of the fastq (that do not exist yet) for the complete run

lukbut (approved):

mwhamgenomics (approved):

analysis_driver/pipelines/demultiplexing.py
```diff
@@ -323,10 +323,6 @@ def _run(self):
                 # make sure the directory where the bam and qc files will go exists
                 makedirs(dirname(self.bam_path(run_element)), exist_ok=True)
                 makedirs(dirname(self.final_fastq_base(run_element)), exist_ok=True)
-                # get the reference genome
```
lukbut: Is there no need to get the reference genome at this stage anymore?

lukbut (approved):

analysis_driver/pipelines/demultiplexing.py
mwhamgenomics: `RunDataset` now implements this as a method - see the `self.dataset.reference_genome(run_element)` on line 329

mwhamgenomics (approved):

Issue #231, opened 2017-03-15
test and integrate bcl2fastq2.20


tcezard:
Download bcl2fast 2.19 from
https://support.illumina.com/sequencing/sequencing_software/bcl2fastq-conversion-software.html
The basic behaviour seems but the stats seems to have changed. That might require some change in the parsers

tcezard:
bcl2fastq v2.17 has a bug where the ConservationStats.xml does not report the accurate number of base in the yield. (see validation report for detail)
Check that new version has this fixed.

tcezard:
bcl2fastq 2.20 is available and we should test this one instead of bcl2fastqv2.19 
http://emea.support.illumina.com/downloads/bcl2fastq-conversion-software-v2-20.html

tcezard:
bcl2fastq stage 
https://github.com/EdinburghGenomics/Analysis-Driver/blob/0c78342d84ab158fc22da88dfe5505353d79c6f1/analysis_driver/pipelines/demultiplexing.py#L71

bcl2fastq command
https://github.com/EdinburghGenomics/Analysis-Driver/blob/0c78342d84ab158fc22da88dfe5505353d79c6f1/analysis_driver/util/bash_commands.py#L11

lukbut:
This issue will be resolved with this PR https://github.com/EdinburghGenomics/Analysis-Driver/pull/369

PR #369, opened 2018-12-19
231 bcl2fastq2.20

lukbut:
This PR includes the changes introduced following the upgrade of bcl2fastq to v2.20.


lukbut:
Closes #231

tests/test_reader/test_demultiplexing_parsers.py
```diff
@@ -8,33 +8,59 @@
...
+        self.assertCountEqual(barcodes,expected_barcode_per_lane)
+        self.assertCountEqual(unknowns,expected_unknown_barcodes_per_lanes)
+
+        self.assertCountEqual(adapter_trimmed_by_id, {
+            'test_run_id_1_ATTACTCG': {'read_1_trimmed_bases': 1056218, 'read_2_trimmed_bases': 1080753},
+            'test_run_id_1_TCCGGAGA': {'read_1_trimmed_bases': 974374, 'read_2_trimmed_bases': 992558},
+            'test_run_id_1_unknown': {'read_1_trimmed_bases': 100051, 'read_2_trimmed_bases': 122287},
+            'test_run_id_2_ATTACTCG': {'read_1_trimmed_bases': 866438, 'read_2_trimmed_bases': 914963},
+            'test_run_id_2_TCCGGAGA': {'read_1_trimmed_bases': 797469, 'read_2_trimmed_bases': 843785},
+            'test_run_id_2_unknown': {'read_1_trimmed_bases': 112371, 'read_2_trimmed_bases': 143299}
```
mwhamgenomics, line 40: `assertCountEqual` doesn't pick up differences in dict values (it casts the observed and expected arguments to lists, giving just the dict keys), so the build still passes if the numbers are changed. `assertEqual` would be a better check here

tests/test_reader/test_demultiplexing_parsers.py
```diff
@@ -8,33 +8,59 @@
...
+            ('2', 'unknown', '485960620')
         ]
-        barcodes, unknowns, barcodeless = dm.parse_conversion_stats(conversion_stat, has_barcode=False)
-        assert barcodeless == expected_barcodeless_per_lane
+        barcodes, unknowns, adapter_trimmed_by_id = dm.parse_json_stats(json_data, 'test_run_id')
+        self.assertCountEqual(barcodes, expected_barcode_per_lane)
+        self.assertCountEqual(unknowns, expected_unknown_barcodes_per_lanes)
+
+        self.assertCountEqual(adapter_trimmed_by_id, {
+            'test_run_id_1': {'read_1_trimmed_bases': 358146231, 'read_2_trimmed_bases': 389554370},
```
mwhamgenomics, line 61: Same here

analysis_driver/reader/demultiplexing_parsers.py
```diff
@@ -2,139 +2,146 @@
 from itertools import islice
```
mwhamgenomics: This and `ElementTree` are no longer used and can be removed

analysis_driver/reader/demultiplexing_parsers.py
```diff
@@ -2,139 +2,146 @@
...
-    all_barcodeless = []
-
-    if not has_barcode:
-        for project in tree.iter('Project'):
-            if project.get('name') == 'all':
-                continue
+def parse_json_stats(json_data, run_id):
+    """Creates an array of tuples of run elements, unknown run elements and associated metadata."""
+    all_run_elements = []
+    unknown_run_elements = []
```
mwhamgenomics: Unused variable - gets reassigned below

mwhamgenomics (dismissed):
AdapterTrimming.txt and ConversionStats.xml in `tests/assets/test_crawlers` have been moved and prefixed with a `_`. These can be removed now that we're using Stats.json instead

analysis_driver/reader/demultiplexing_parsers.py
```diff
@@ -2,139 +2,146 @@
...
+        # obtaining number of trimmed bases and bases r1 and r2 q30, confirming the read number first
+        nb_bases_r1_q30, nb_bases_r2_q30 = 0, 0
+        if r1['ReadNumber'] == 1 and r2['ReadNumber'] == 2:
+            nb_bases_r1_q30 = r1['YieldQ30']
+            nb_bases_r2_q30 = r2['YieldQ30']
+            trimmed_bases_r1 = r1['TrimmedBases']
+            trimmed_bases_r2 = r2['TrimmedBases']
+        elif r1['ReadNumber'] == 2 and r2['ReadNumber'] == 1:
+            # this is not expected
+            raise NotImplementedError()
```
mwhamgenomics: NotImplementedError is [intended for defining abstract methods](https://docs.python.org/3/library/exceptions.html?highlight=notimplementederror#NotImplementedError) - I'd raise something like `SequencingRunError`. Not too important right now, though - can refactor later

analysis_driver/util/helper_functions.py
```diff
@@ -0,0 +1,6 @@
+# retrieving run_element_id, which differs between barcodeless and multiplexed runs
+def get_run_element_id(run_id, lane_number, barcode=None):
+    if barcode is None:
+        return '{dataset_name}_{lane}'.format(dataset_name=run_id, lane=lane_number)
+    else:
+        return '{dataset_name}_{lane}_{barcode}'.format(dataset_name=run_id, lane=lane_number, barcode=barcode)
```
mwhamgenomics, line 6: It's annoying that we need to do this, but it all works fine

analysis_driver/reader/demultiplexing_parsers.py
lukbut: @mwhamgenomics I agree, that's why I placed it in a helper function so at least it's only found once in the source code.

tests/test_reader/test_demultiplexing_parsers.py
lukbut: @mwhamgenomics I was using assertEqual before I used assertCountEqual, but it turns out assertEqual checks the order of the elements in the dict too, which was not always deterministic and caused issues. I'll try to find a better solution.

analysis_driver/reader/demultiplexing_parsers.py
lukbut: Removed

analysis_driver/reader/demultiplexing_parsers.py
lukbut: Removed.

analysis_driver/util/helper_functions.py
lukbut: @mwhamgenomics I agree, that's why I placed it in a helper function so at least it's only found once in the source code.

analysis_driver/reader/demultiplexing_parsers.py
lukbut: @mwhamgenomics Technically this isn't an error. There would be nothing wrong if the first ReadNumber is actually the R2 value, and the second ReadNumber is the R1 value. The code to handle this flipped version has not been included yet, as I was not sure whether this could ever happen, so I thought it would be best to handle the case and raise a NotImplementedError(). In the eventuality of it being raised, we can implement it.

tests/test_reader/test_demultiplexing_parsers.py
lukbut: @mwhamgenomics As discussed, this seemed to be the case when comparing lists of tuples with assertEqual. Comparisons of lists of tuples are now being made using assertCountEqual.

analysis_driver/reader/demultiplexing_parsers.py
```diff
@@ -1,140 +1,144 @@
...
+
+    return all_run_elements, unknown_run_elements, adapter_trimmed_by_id
+
+
+def _parse_conversion_results_from_json_stats(run_id, lane, all_run_elements, adapter_trimmed_by_id):
+    for sample in lane['DemuxResults']:
+        """ If the run is not multiplexed, the index_sequence will not be present in the JSON file, and the cluster
+        count needs to be retrieved from a different variable in the JSON file. The index_sequence variable will be 
+        prepopulated as None. """
+        r1, r2 = sample['ReadMetrics'][0], sample['ReadMetrics'][1]
```
tcezard:
```suggestion
        r1 = sample['ReadMetrics'][0] if sample['ReadMetrics'][0]['ReadNumber'] == 1 else sample['ReadMetrics'][1]
        r2 = sample['ReadMetrics'][1] if sample['ReadMetrics'][1]['ReadNumber'] == 2 else sample['ReadMetrics'][0]
```
an no need for the `NotImplementedError` in L60

analysis_driver/reader/demultiplexing_parsers.py
```diff
@@ -1,140 +1,144 @@
...
+        if r1['ReadNumber'] == 1 and r2['ReadNumber'] == 2:
+            nb_bases_r1_q30 = r1['YieldQ30']
+            nb_bases_r2_q30 = r2['YieldQ30']
+            trimmed_bases_r1 = r1['TrimmedBases']
+            trimmed_bases_r2 = r2['TrimmedBases']
+        elif r1['ReadNumber'] == 2 and r2['ReadNumber'] == 1:
+            # this is not expected
+            raise NotImplementedError()
+
+        if r1['Yield'] != r2['Yield']:
```
tcezard: I would prefer 
```python
        assert r1['Yield'] == r2['Yield'], "yield is expected to be equal for r1 and r2"
```
It is shorter and as descriptive

etc/tool_versioning.yaml
```diff
@@ -31,6 +31,21 @@ toolsets:
...
+
+            fastq_filterer: null
+
+            fastqc: null
+
+            pigz: null
+
+            seqtk: null
+
+            well_duplicates: null
```
tcezard, line 49: Can you add a comment describing how to add tools that enherit from the previous toolset.
I had to go back to the code to understand how it's working

tcezard (dismissed):

etc/tool_versioning.yaml
mwhamgenomics: There's a test config in `tests/assets/tool_versioning.yaml` with (admittedly brief) descriptions of how the inheritance works

analysis_driver/reader/demultiplexing_parsers.py
lukbut: Thanks @tcezard I'll implement this.

etc/tool_versioning.yaml
lukbut: Sure, I'll add this.

analysis_driver/reader/demultiplexing_parsers.py
lukbut: @tcezard I am still leaving the error, but converting it to a value error, just in case this doesn't work -it should never not work, but I'm catering for the eventuality just the same.

analysis_driver/reader/demultiplexing_parsers.py
lukbut: I've change this as instructed - I agree it is more appropriate.

tcezard (approved):

PR #375, opened 2019-02-06
Java versioning, fixes and cleanup

### LINKED ISSUE #279: Version of java tools cannot be determined in cron
When extracting the version of GATK, Fastqc, or RTG, AnalysisDriver need access to java application.
When running it on a terminal java is provided by the environment (`.bashrc`) but when running from cron, java is not available hence the version cannot be resolved.
A workaround can be implemented to add the java environment in cron by preceding the command with a `source`, however a better solution might be found.

This also highlighted that:
 - Integration tests could not pick this up as they do not run on cron
 - When tool versioning fails it does not mark the process as fail as it was never started. (tools version resolution happens before `dataset.start`)

### LINKED ISSUE #345: Project process: batch samples when more than 30 are processed together
Use CombineGVCFs to create batch of GVCF until you have less than 30 then run GenotypeGVCF.

### LINKED ISSUE #355: Config cleanup
- [x] At the moment, data output uses the config `sample -> delivery_dest`, when it could just use `delivery -> dest` like everything else.
- [x] We might be able to remove `delivery_source` and use `sample -> output_dir`
- [x] It might also be a good idea to remove the config merging

### LINKED ISSUE #358: Trio checking for projects with more than 25 samples
At the moment `ProjectDataset.samples_processed` doesn't depaginate, so projects with more than 25 samples don't get picked up without being forced.

### LINKED ISSUE #365: Spam reduction
At the moment, the messages that the pipeline logs are very verbose. We should review the logging and notifications, remove stuff we don't use and potentially add things we want to see.

mwhamgenomics:
- Removing config merging - closes #355 
- Removing `--debug` option, adding debug log in job folder, setting repo log to `INFO`, rejigging some log message levels - closes #365 
  - [x] awaiting validation
- Depaginating ProjectDataset.samples_processed - closes #358 
- Allowing versioned tools to depend on another versioned tool (e.g. Picard depending on Java) - closes #279 
- Removing toolset patching - example_analysisdriver.yaml now includes all versioned tools (in alphabetical order!)


tests/test_bash_commands.py
```diff
@@ -3,207 +3,184 @@
 from analysis_driver.util import bash_commands
 from analysis_driver.reader import RunInfo
 
-assets = TestAnalysisDriver.assets_path
-fastq_path = TestAnalysisDriver.fastq_path
-
-
-def test_bcl2fastq():
```
mwhamgenomics, line 10: These tests had to be moved inside a TestAnalysisDriver object so the config and toolset would be loaded properly.

analysis_driver/dataset.py
```diff
@@ -116,15 +116,16 @@ def soft_terminate(self):
...
-            self.debug('Attempted to terminate invalid pid %s with signal %s', pid, signal_id)
+            self.warning('Termination unsuccessful')
             return
 
         os.kill(pid, signal_id)
         while self._pid_running(pid):
             sleep(1)
-        self.info('Terminated pid %s for %s %s with signal %s', pid, self.type, self.name, signal_id)
+
+        self.info('Termination successful')
```
tcezard, line 128: None of these comment are ever seen because at the point of calling terminate we have not setup the file logging and I don't think stdout is setup either

analysis_driver/pipelines/projects.py
```diff
@@ -39,6 +52,34 @@ def build_pipeline(dataset):
     return cleanup
 
 
+class CombineGVCFs(segmentation.Stage):
+    gvcfs = segmentation.ListParameter()
+    output_base = segmentation.Parameter()
+    reference = segmentation.Parameter()
```
tcezard: Is the `reference` parameter required since `CombineGVCFs` has access to `dataset`?

tests/test_pipelines/test_projects.py
```diff
@@ -3,7 +3,6 @@
 from unittest.mock import patch
 from tests.test_analysisdriver import TestAnalysisDriver, NamedMock
 from analysis_driver.pipelines import projects
-from analysis_driver.exceptions import PipelineError
 
 test_projects = os.path.join(TestAnalysisDriver.assets_path, 'test_projects')
 relatedness_outfiles = os.path.join(test_projects, 'relatedness_outfiles')
```
tcezard, line 8: An extra test here for build_pipeline would be good. It should mock many samples and test that the combining in batches is setup as expected.

tcezard (changes requested):

analysis_driver/client.py
```diff
@@ -20,25 +20,17 @@ def main(argv=None):
...
-            cfg.merge(cfg['run'])
-        scanner = RunScanner(cfg)
+        scanner = RunScanner()
     elif args.sample:
-        if 'sample' in cfg:
-            cfg.merge(cfg['sample'])
-        scanner = SampleScanner(cfg)
+        scanner = SampleScanner()
     else:
         assert args.project
```
lukbut: Why not just use another elif, and then raise an Exception in a further else if args does not fit into any case? Just to maintain the flow of code logic design. Of course this is down to personal preference.

analysis_driver/pipelines/demultiplexing.py
```diff
@@ -24,7 +24,6 @@ def fastq_dir(self):
 
 class Setup(DemultiplexingStage):
     def _run(self):
-        self.info('Job dir: ' + self.job_dir)
```
lukbut, line 30: Why has this been removed?

analysis_driver/quality_control/bcl_validation.py
```diff
@@ -117,7 +117,7 @@ def run_bcl_check(self, bcls, slice_size=100, max_job_number=500):
             if os.path.isfile(validation_log_tmp):
                 for bcl, exit_status in self.read_check_bcl_files(validation_log_tmp):
                     f.write('%s,%s\n' % (bcl, exit_status))
-        self.info('Finished validation. Found %s invalid files' % len(self.read_invalid_files()))
+        self.info('Finished validation. Found %s invalid files', len(self.read_invalid_files()))
```
lukbut, line 113: Was this a bug?

analysis_driver/tool_versioning.py
```diff
@@ -27,49 +28,67 @@ def select_version(self, version):
...
-                self.tools[k] = cfg['tools'][k]
+            self.add_tool(k)
 
         self.info('Selected %s toolset version %s', self.type, self.version)
 
     @property
     def unversioned_tools(self):
         return [k for k in self.tools if k not in self.tool_versions]
 
+    def add_tool(self, toolname):
```
lukbut, line 39: Great

etc/example_analysisdriver.yaml
```diff
@@ -37,31 +36,32 @@ default:
...
         tabix: path/to/tabix
-        fastqscreen: path/to/fastqscreen
+        vcftools: path/to/vcftools
         verifybamid: path/to/verifybamid
         well_duplicates: path/to/well_duplicates
-        fastq-filterer: path/to/fastq-filterer
-        vcftools: path/to/vcftools
-        blastn: path/to/blastn
-        peddy: path/to/peddy
-        fix_dup_unmapped: path/to/fix_dup_unmapped
```
lukbut, line 64: So are all the removed programs here, programs which are no longer used in analysis_driver?

tests/assets/tool_versioning.yaml
```diff
@@ -26,9 +31,13 @@ toolsets:
...
             bwa:  # inherit version_cmd from previous version
                 version: 1.1
 
-            samtools: null  # inherit all
\ No newline at end of file
+            samtools: null  # inherit all
+            java: null
+
+    fake_tools:
+        0: {}
```
lukbut: No new line at the end of file.

tests/test_tool_versioning.py
```diff
@@ -5,33 +5,34 @@
...
 
         self.toolset.select_version(1)
         assert self.toolset['bwa'] == 'path/to/bwa_1.1'
-        assert self.toolset['gatk'] == 'path/to/GenomeAnalysisTK.jar_4'
+        assert self.toolset['gatk'] == 'path/to/gatk_4'
         assert self.toolset['fastqc'] == 'path/to/fastqc_v0.11.5'
+        assert self.toolset['java'] == 'path/to/java_8'
         assert 'fastqc' in self.toolset.unversioned_tools
-        assert self.toolset.tool_versions == {'bwa': 1.1, 'gatk': 4, 'samtools': '1.3.1'}
+        assert self.toolset.tool_versions == {'bwa': 1.1, 'gatk': 4, 'samtools': '1.3.1', 'java': 8}
```
lukbut, line 41: It might be worth adding picard here too?

analysis_driver/pipelines/demultiplexing.py
mwhamgenomics: This has been moved to client.py, line 106

analysis_driver/pipelines/projects.py
mwhamgenomics: Yes, looks like we can just reference `self.dataset.reference_genome`.

In that case, we can also remove the parameter from GenotypeGVCFs - unless we want to align to a different genome in certain cases, for some reason? Can't imagine us doing that at the moment, though.

Edit: done.

analysis_driver/quality_control/bcl_validation.py
mwhamgenomics: This is the recommended way of doing string formatting in log messages. The way `logging` handles this, formatting is only performed if the message is actually logged. I guess it's intended to avoid wasting CPU time formatting thousands of debug messages when you only have logging set to `INFO` or above (for example).

etc/example_analysisdriver.yaml
mwhamgenomics: This has been put into alphabetical order, so things have only been moved around, and `fastq-filterer` has been renamed to `fastq_filterer` so it fits tool_versioning.yaml

analysis_driver/quality_control/bcl_validation.py
lukbut: Oh I see, so it's not calculated unless necessary.

tests/test_tool_versioning.py
mwhamgenomics: Not really - it doesn't matter what the tools/toolsets are called here, we're only testing the inheritance of versions and commands.

lukbut: Right, for a second I thought you were testing the contents of the file.

analysis_driver/dataset.py
mwhamgenomics: Have added a stdout logging handler to the manual commands (reset, skip, stop, soft stop, etc.)

tests/test_pipelines/test_projects.py
mwhamgenomics: Implemented

mwhamgenomics:
After some investigation of GenotypeGVCFs and CombineGVCFs on genomic data, we have found the following (all in 4 threads unless specified otherwise):
- on a project of 77 files, genotyping failed in ~4 days, 5 hrs (231gb allocated memory)
- on 57 files, genotyping succeeded in 2 days, 19 hrs (171gb)
- on 37 files, genotyping succeeded in ~15 hrs (111gb)
  - combining and genotyping succeeded in 3 days, 9 hrs
  - genotyping in 12 threads succeeded in 6 hrs
- on 48 files (for which trio checking succeeded in production), genotyping failed in 2 days, 13 hrs (144 gb)
  - combining and genotyping is still running after 4 days

In all cases where GenotypeGVCFs ran out of memory, it proceeded as normal until either chrY or chrEBV, where it would get stuck at one locus. The estimated remaining time would then increase, and eventually GATK would fail with a `java.lang.OutOfMemoryError`.

Even if CombineGVCFs fixes the trio check memory problems, the overhead of merging makes this impractical to run on everything in a static pipeline. Two potential options are:
- do not proceed with CombineGVCFs
- make the pipeline dynamic, and only run CombineGVCFs if there are more than `x` samples

We will keep issue #345 open, revert the CombineGVCFs changes here, and note in the issue that this would need to be either a dynamic pipeline or an implementation in GATK4 (see #377)

lukbut:
@mwhamgenomics May I ask why an implementation in GATK4 will help?

lukbut (approved):

mwhamgenomics:
The behaviour seen with GATK 3 is unexpected given that it always seems to occur on small chromosomes, and it is possible that GATK4 fixes this. It's also unlikely that the Broad Institute would be willing to help us with the version we're currently using.

etc/tool_versioning.yaml
```diff
@@ -86,7 +86,8 @@ toolsets:
```
tcezard, line 87: picard in line 46 will also need a dependency to java.

tcezard (changes requested):

tcezard (approved):

PR #378, opened 2019-02-13
Fix early alignment when a run is aborted

tcezard:
This PR fixes an issue introduced when combining early alignment #368 and support for aborted runs #362.
It will allow the pipeline to finish with no error in the event of an aborted run.

### DIFF
--- a/analysis_driver/pipelines/demultiplexing.py
+++ b/analysis_driver/pipelines/demultiplexing.py
@@ -242,6 +242,13 @@ class WaitForRead2(DemultiplexingStage):
         while current_cycle < required_number_of_cycle and self.dataset.is_sequencing():
             time.sleep(1200)
             current_cycle = sorted(interop_metrics.get_cycles_extracted(self.dataset.input_dir))[-1]
+
+        # make sure the run is not aborted or errored before continuing with the rest of the pipeline
+        run_status = self.dataset.lims_run.udf.get('Run Status')
+        if run_status not in ['RunCompleted', 'RunStarted', 'RunPaused']:
+            self.error('Run status is \'%s\'. Stopping.', run_status)
+            raise SequencingRunError(run_status)
+
         return 0
 
 
--- a/integration_tests/integration_test.py
+++ b/integration_tests/integration_test.py
@@ -211,11 +211,12 @@ class IntegrationTest(ReportingAppIntegrationTest):
                 ),
                 self.cfg['demultiplexing']['lane_qc']
             )
-        self.expect_stage_data(['setup', 'wellduplicates', 'bcl2fastq', 'phixdetection', 'fastqfilter', 'seqtkfqchk',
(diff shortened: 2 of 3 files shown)


integration_tests/integration_test.py
```diff
@@ -211,11 +211,12 @@ def test_demultiplexing(self):
...
                 self.cfg['demultiplexing']['lane_qc']
             )
-        self.expect_stage_data(['setup', 'wellduplicates', 'bcl2fastq', 'phixdetection', 'fastqfilter', 'seqtkfqchk',
-                               'md5sum', 'fastqc', 'integritycheck', 'qcoutput1', 'dataoutput', 'cleanup',
-                               'samtoolsdepthmulti', 'picardinsertsizemulti', 'qcoutput2', 'runreview',
-                               'picardmarkduplicatemulti', 'samtoolsstatsmulti', 'bwaalignmulti', 'waitforread2',
-                               'bcl2fastqhalfrun', 'picardgcbias'])
+        self.expect_stage_data([
+            'setup', 'wellduplicates', 'bcl2fastq', 'phixdetection', 'fastqfilter', 'seqtkfqchk', 'md5sum', 'fastqc',
+            'integritycheck', 'qcoutput1', 'dataoutput', 'cleanup', 'samtoolsdepthmulti', 'picardinsertsizemulti',
```
lukbut, line 216: Can we name the stages in a more informative way than "qc_output_1"?

analysis_driver/pipelines/demultiplexing.py
```diff
@@ -242,6 +242,13 @@ def _run(self):
         while current_cycle < required_number_of_cycle and self.dataset.is_sequencing():
             time.sleep(1200)
             current_cycle = sorted(interop_metrics.get_cycles_extracted(self.dataset.input_dir))[-1]
+
+        # make sure the run is not aborted or errored before continuing with the rest of the pipeline
+        run_status = self.dataset.lims_run.udf.get('Run Status')
+        if run_status not in ['RunCompleted', 'RunStarted', 'RunPaused']:
+            self.error('Run status is \'%s\'. Stopping.', run_status)
+            raise SequencingRunError(run_status)
```
mwhamgenomics, line 250: A test for this `raise` would be good

mwhamgenomics (changes requested):

integration_tests/integration_test.py
tcezard: That would be done in a different PR but we could discuss alternative naming conventions.
However we want the stage name to be short and readable because they appear on the reporting app.

analysis_driver/pipelines/demultiplexing.py
tcezard: Test have now been added

mwhamgenomics (approved):

PR #383, opened 2019-03-12
start Sample processing using coverage

### LINKED ISSUE #374: SampleProcess start criteria should include coverage
The Sample processing pipeline only starts when the required yield is meeting the expected yield.
We often need to force sample to start because they do not meet the required yield but have enough coverage.
We should change the criteria to be 
 - yield > required yield
or
 - coverage > required coverage

tcezard:
Add support for looking up coverage calculated in the run elements
rename data_threshold to required_yield_threshold
closes #374

### DIFF
--- a/analysis_driver/dataset.py
+++ b/analysis_driver/dataset.py
@@ -429,12 +429,11 @@ class SampleDataset(Dataset):
     endpoint = 'samples'
     id_field = 'sample_id'
 
-    def __init__(self, name, most_recent_proc=None, data_threshold=None):
+    def __init__(self, name, most_recent_proc=None):
         super().__init__(name, most_recent_proc)
         self._run_elements = None
         self._sample = None
         self._non_useable_run_elements = None
-        self._data_threshold = data_threshold
         self._species = None
         self._genome_version = None
         self._user_sample_id = None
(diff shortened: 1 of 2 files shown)


lukbut (approved):

analysis_driver/dataset.py
```diff
@@ -529,14 +535,18 @@ def _pipeline_instruction(self):
...
+            (self.required_coverage_threshold and self._amount_coverage() > self.required_coverage_threshold)
+        )
 
     def report(self):
         runs = query_dict(self.sample, 'aggregated.run_ids')
         non_useable_runs = sorted(set(r[ELEMENT_RUN_NAME] for r in self.non_useable_run_elements))
 
-        s = '%s  (%s / %s  from %s) ' % (
-            super().report(), self._amount_data(), self.data_threshold, ', '.join(runs)
+        s = '%s  (%s / %s and %s / %s from %s) ' % (
```
mwhamgenomics: It's not immediately obvious in the report which number is which:

    test_dataset -- this, that, other (1500000000 / 1000000000 and 32 / 30 from a_run, another_run)

I think it would be good to label these numbers, e.g. `yield 1500000000 / 1000000000 and X coverage 32 / 30`, or something similar like adding units.

tcezard: I've added `yield` and `coverage` prefix

mwhamgenomics (approved):

tcezard:
Validation ID: VB8

PR #384, opened 2019-03-15
Check the existence of bcls

### LINKED ISSUE #382: Searching for current cycle when they do not exist yet breaks
We should check that the cycles exist then take the last one or return 0
see:
https://github.com/EdinburghGenomics/Analysis-Driver/blob/096af2f460317f81de749c5de256b8e338a9b83b/analysis_driver/pipelines/demultiplexing.py#L241

tcezard:
Check that the bcl exists before the returning the number of existing cycles
closes #382

### DIFF
--- a/analysis_driver/quality_control/interop_metrics.py
+++ b/analysis_driver/quality_control/interop_metrics.py
@@ -121,8 +121,9 @@ def get_last_cycles_with_existing_bcls(run_dir):
     run_metrics = RunMetrics()
     run_metrics.read(run_dir)
     extraction_metrics = run_metrics.extraction_metric_set()
-    all_cycles = extraction_metrics.cycles()
+    all_cycles = sorted(extraction_metrics.cycles())
     last_complete_cycles = 0
+    # start from the last cycle and walk back until found a cycle with all the bcl present.
     for cycle in all_cycles[::-1]:
         all_bcls = glob.glob(os.path.join(run_dir, 'Data/Intensities/BaseCalls/L00*/C%s.1/*.bcl.gz' % cycle))
         if len(all_bcls) == extraction_metrics.metrics_for_cycle(cycle).size():
--- a/analysis_driver/quality_control/interop_metrics.py
+++ b/analysis_driver/quality_control/interop_metrics.py
@@ -114,7 +114,7 @@ class BadTileCycleDetector(AppLogger):
 def get_last_cycles_with_existing_bcls(run_dir):
     """
     This function checks the extracted cycle from the interop and confirm the presence of the bcl files on the filesystem.
-    the confirmation is only performed from the last cycles and the first full confirmed cycle is returned.
+    The confirmation is only performed from the last cycles and the first full confirmed cycle is returned.
     :param run_dir: The location where the run is stored
     :returns: the last cycle of the run with existing bcls
     """
--- a/analysis_driver/quality_control/interop_metrics.py
+++ b/analysis_driver/quality_control/interop_metrics.py
@@ -113,7 +113,7 @@ class BadTileCycleDetector(AppLogger):
 
 def get_last_cycles_with_existing_bcls(run_dir):
     """
-    This function check the extracted cycle from the interop and confirm the presence of the bcl files o the filesystem.
+    This function checks the extracted cycle from the interop and confirm the presence of the bcl files on the filesystem.
     the confirmation is only performed from the last cycles and the first full confirmed cycle is returned.
     :param run_dir: The location where the run is stored
     :returns: the last cycle of the run with existing bcls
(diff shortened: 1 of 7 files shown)


analysis_driver/quality_control/interop_metrics.py
```diff
@@ -109,9 +111,22 @@ def detect_bad_cycles(self):
         return bad_cycle_per_lanes
 
 
-def get_cycles_extracted(run_dir):
+def get_last_cycles_with_existing_bcls(run_dir):
+    """
+    This function check the extracted cycle from the interop and confirm the presence of the bcl files o the filesystem.
```
lukbut:
```suggestion
    This function checks the extracted cycle from the interop and confirm the presence of the bcl files on the filesystem.
```

analysis_driver/quality_control/interop_metrics.py
```diff
@@ -109,9 +111,22 @@ def detect_bad_cycles(self):
...
+    the confirmation is only performed from the last cycles and the first full confirmed cycle is returned.
```
lukbut:
```suggestion
    The confirmation is only performed from the last cycles and the first full confirmed cycle is returned.
```

analysis_driver/quality_control/interop_metrics.py
```diff
@@ -109,9 +111,22 @@ def detect_bad_cycles(self):
...
+    :returns: the last cycle of the run with existing bcls
+    """
     run_metrics = RunMetrics()
     run_metrics.read(run_dir)
     extraction_metrics = run_metrics.extraction_metric_set()
-    return sorted(extraction_metrics.cycles())
+    all_cycles = extraction_metrics.cycles()
+    last_complete_cycles = 0
+    for cycle in all_cycles[::-1]:
+        all_bcls = glob.glob(os.path.join(run_dir, 'Data/Intensities/BaseCalls/L00*/C%s.1/*.bcl.gz' % cycle))
```
lukbut: Just to confirm, you are checking the number of BCLs in each cycle going backwards from the last cycle, and comparing it with the one you are extracting from the metrics, and then returning the last completed cycle (which is the one whose number of BCLs match the number extracted from the metrics.

tests/test_pipelines/test_demultiplexing.py
```diff
@@ -107,8 +107,8 @@ def test_run(self):
                             lims_run=Mock(udf={'Run Status': 'RunStarted'}))
 
         # cycle extracted states 310 cycles done
-        pcycles = patch('analysis_driver.quality_control.interop_metrics.get_cycles_extracted',
-                        return_value=range(1, 311))
+        pcycles = patch('analysis_driver.quality_control.interop_metrics.get_last_cycles_with_existing_bcls',
+                        return_value=310)
```
lukbut, line 111: Why 310 not 311? Is it because you are now compensating in the function?

lukbut (commented):
I have gone through the code and have the following comments, but I do not feel I can vouch for the changes in technical functionality as I am not familiar with this part of the code at all.

tests/test_pipelines/test_demultiplexing.py
tcezard: The previous function returned a list of cycle which is mimicked using `range`. in `range` the end is not inclusive.
The new function returns only the last cycle value.

analysis_driver/quality_control/interop_metrics.py
tcezard: That's right. I've added a comment to clarify

analysis_driver/quality_control/interop_metrics.py
```diff
@@ -109,9 +111,22 @@ def detect_bad_cycles(self):
...
     run_metrics = RunMetrics()
     run_metrics.read(run_dir)
     extraction_metrics = run_metrics.extraction_metric_set()
-    return sorted(extraction_metrics.cycles())
-
+    all_cycles = sorted(extraction_metrics.cycles())
+    last_complete_cycles = 0
+    # start from the last cycle and walk back until found a cycle with all the bcl present.
+    for cycle in all_cycles[::-1]:
+        all_bcls = glob.glob(os.path.join(run_dir, 'Data/Intensities/BaseCalls/L00*/C%s.1/*.bcl.gz' % cycle))
```
mwhamgenomics: Ideally we'd want to be consistent with the rest of the repo and use `egcg_core.util.find_files`:

```python
util.find_files(run_dir, 'Data', 'Intensities', 'BaseCalls', 'L00*', 'C%s.1' % cycle, '*.bcl.gz')
```

But this should be okay until we port the pipeline to MacOS 9 😐.

mwhamgenomics (approved):

analysis_driver/quality_control/interop_metrics.py
tcezard: I've changed it to `util.find_files(run_dir, ...`

tcezard:
This PR will not go through validation as it address an urgent bug. The change does not modify any feature that can influence the data or the metrics but only when the data will start being processed.

PR #386, opened 2019-03-18
Make the samplesheet generation deterministic

### LINKED ISSUE #385: non deterministic samplesheet
When bcl2fastq is run multiple time accros multiple run, the samplesheet is generated several times and the sample can be ordered differently in the different version.
That can result in inconsitent naming in the fastq files and/or duplicate fastq.
We should either:
 - make the order of the run element deterministic
 - load the samplesheet files if it exists rather than generate it

tcezard:
closes #385

### DIFF
--- a/analysis_driver/dataset.py
+++ b/analysis_driver/dataset.py
@@ -293,7 +293,8 @@ class RunDataset(Dataset):
     def sample_sheet_file(self):
         if self._sample_sheet_file is None:
             self._sample_sheet_file = os.path.join(self.input_dir, 'SampleSheet_analysis_driver.csv')
-            self._generate_samplesheet(self._sample_sheet_file)
+            if not os.path.isfile(self._sample_sheet_file):
+                self._generate_samplesheet(self._sample_sheet_file)
         return self._sample_sheet_file
 
     def _generate_samplesheet(self, filename):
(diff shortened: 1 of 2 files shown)


analysis_driver/dataset.py
```diff
@@ -293,7 +293,8 @@ def run_info(self):
     def sample_sheet_file(self):
         if self._sample_sheet_file is None:
             self._sample_sheet_file = os.path.join(self.input_dir, 'SampleSheet_analysis_driver.csv')
-            self._generate_samplesheet(self._sample_sheet_file)
+            if not os.path.isfile(self._sample_sheet_file):
```
lukbut, line 296: @tcezard So the file is no longer generated if it already exists? Could this be a problem if we need to repeat the sample sheet file generation and it already exists?

tcezard: There might be cases in the future when we will need to remove an erroneous samplesheet to make sure it gets regenerated.

lukbut: May I ask why we are not regenerating every time this function is called, as was happening previously?

tcezard: We could now that the generation is deterministic but there is no need. The Samplesheet is not suppose to change

tcezard:
This PR will not go through validation as it address an urgent bug. The change does not modify any feature that can influence the data or the metrics but only the order in which the line appear in the samplesheet.

mwhamgenomics (approved):

Issue #274, opened 2017-08-16
Generic DataOutput


katieemelianova:
Data output is currently coded in projects, run processing and in common as SampleDataOutput - these could be rewritten to be more generic.

mwhamgenomics:
Since we MD5 every file we output, this should also include MD5Sum.

PR #388, opened 2019-03-25
Rapid analysis

### LINKED ISSUE #372: Rapid processing pipeline
This will use Dragen to process a non-pooled run from BCL to Bam/VCF:

- This will be in addition to normal run/sample processing. The slow data will be used for QC and comparison.
- Top-ups as introduced in #69 will not be supported - this would defeat the point of rapid processing.
- Data will be processed and delivered with minimal review, none of which will be manual.
  - If automated checks fail, then we stop.

The pipeline will need to:

- [x] Wait until sequencing is complete
- [x] Pick up a signal that rapid processing should be run on whatever lanes in the run.
  - Lims UDF from samples via RunDataset._run_elements_from_lims
  - We may have to merge lanes later for 60/90X
- [x] Check the Dragen usage quota to ensure that it can be run
- [x] Write a sample sheet compatible with Dragen
- [x] Run Dragen using `job_queue=...` in executor
  - [x] Will need to check whether Dragen can run multiple processes at once. If not, then figure out how to enforce that only one Dragen job runs at a time.
  - We may need to monitor Dragen licence usage - manually or via executor
- [x] Extract run and sample QC from Dragen's metrics
- [x] Call automatic_review on Reporting-App to populate `rapid_processing_useable` and `rapid_processing_reviewed`
- [x] If review fails, then stop
  - Fail the pipeline?
- [x] Output rapid data to a separate `rapid_samples` subdir in the project folder to avoid file conflicts
- [x] Also link to delivered_projects as per data delivery
- [x] Notify that samples have finished rapid processing
  - push to a rapid_processing_complete field in `samples` 

This will require the changes in EGCG-Core addressing EdinburghGenomics/EGCG-Core#97.

mwhamgenomics:
- Adds a rapid analysis pipline with [Dragen](https://www.illumina.com/products/by-type/informatics-products/dragen-bio-it-platform.html) - closes #372.
- Removes output config spaghetti
- Integration tests:
  - Adds `grep -v` support to expect_output_files
  - Adds easier dynamic patching. IntegrationTest.patches in EGCG-Core should probably be a list now, not a tuple.


analysis_driver/pipelines/demultiplexing.py
```diff
@@ -486,7 +487,12 @@ def stage(cls, **params):
...
                        previous_stages=[qc_output, stats_output, depth_output, md_output, is_output, gc_bias])
-    data_output = stage(DataOutput, previous_stages=[qc_output2])
-    _cleanup = stage(Cleanup, previous_stages=[data_output])
-    review = stage(RunReview, previous_stages=[_cleanup])
+    data_output = stage(DataOutput, previous_stages=[qc_output, qc_output2])
+    cleanup = stage(Cleanup, previous_stages=[data_output])
+    final_checkpoint = [cleanup]
+
+    if dataset.rapid_samples_by_lane:
+        final_checkpoint.append(rapid._build_pipeline(dataset, setup))
```
tcezard: Don't we want the rapid to happen before the cleanup?
I know it is unlikely since it's much faster but the way it is written at the moment the cleanup could be done before the rapid is completed

analysis_driver/dataset.py
```diff
@@ -344,6 +346,23 @@ def _find_pooling_step_for_artifact(art, expected_pooling_step_name=None, max_it
...
+            for lane, artifact in flowcell.placements.items():
+                if len(artifact.reagent_labels) > 1:
+                    continue  # we don't want to run rapid processing on pools
+
+                assert len(artifact.samples) == 1
+                sample = artifact.samples[0]
+                if sample.udf.get('Rapid Analysis'):
+                    self._rapid_samples_by_lane[lane.split(':')[0]] = sample
+
+        return self._rapid_samples_by_lane
```
tcezard, line 375: Not sure about this function: 
On the one hand it duplicates parts of `_run_elements_from_lims` which is not great and accessing the LIMS via REST API makes it difficult to test. 
On the other it's probably the simplest way to access the sample UDF.
The alternative would be to query `get_document('sample_info', match={"sample_id": sample_id})` for each sample. 
Also I'm not too keen having `pyclarity_lims` object being carried around in Analysis driver.

analysis_driver/pipelines/rapid.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+class Dragen(RapidStage):
+    def _run(self): 
+        cmds = []
+        for lane, sample in self.dataset.rapid_samples_by_lane.items():
+            cmds.append(
+                'if [ -d {tmp_dir} ]; then rm -r {tmp_dir}; fi; mkdir -p {tmp_dir}; {dragen} -r {ref} '
+                '--bcl-input-dir {run_dir} --bcl-only-lane {lane} --output-directory {tmp_dir} '
+                '--output-file-prefix {user_sample_id} --enable-variant-caller true --vc-sample-name {sample_id} '
```
tcezard: I think the `--vc-sample-name` should be `{user_sample_id}`

tcezard (changes requested):
Mostly comments or suggestions

analysis_driver/dataset.py
mwhamgenomics: We'd need to go through the sample sheet to know what samples to query. We'd also need to figure out how to insert fake Lims data in the integration test. Added a Todo for now, and returning a dict so that at least we're returning strings

analysis_driver/pipelines/demultiplexing.py
mwhamgenomics: Fixed

analysis_driver/pipelines/common.py
```diff
@@ -83,19 +83,14 @@ def stage(cls, **params):
 class SampleDataOutput(segmentation.Stage):
     output_fileset = segmentation.Parameter()
 
-    @property
-    def output_cfg(self):
-        output_file_config.set_pipeline_type(self.output_fileset)
-        return output_file_config
```
lukbut, line 89: @mwhamgenomics Where is this happening now?

analysis_driver/pipelines/rapid.py
```diff
@@ -0,0 +1,238 @@
...
+
+class RapidStage(segmentation.Stage):
+    def rapid_output_dir(self, lane):
+        return os.path.join(self.job_dir, 'rapid_analysis_%s' % lane)
+
+
+class Dragen(RapidStage):
+    def _run(self): 
+        cmds = []
+        for lane, sample in self.dataset.rapid_samples_by_lane.items():
```
lukbut, line 20: A little comment would be great here to explain what you're doing.

analysis_driver/pipelines/rapid.py
```diff
@@ -0,0 +1,238 @@
...
+                    ref=cfg['dragen']['reference'], run_dir=self.input_dir, lane=lane,
+                    out_dir=self.rapid_output_dir(lane), user_sample_id=sample['User Sample Name'],
+                    sample_sheet=self.dataset.sample_sheet_file, dbsnp=cfg['dragen']['dbsnp'],
+                    tmp_dir=os.path.join(cfg['dragen']['staging'], self.dataset.name + '_' + lane)
+                )
+            )
+
+        return executor.execute(
+            *cmds,
+            prelim_cmds=['ulimit -n 65535', 'ulimit -c 0', 'ulimit -s 8192', 'ulimit -u 16384', 'ulimit -i 1029522'],
```
lukbut, line 52: How did you get to these ulimit values? It would be best to document this for future reference.

analysis_driver/pipelines/rapid.py
```diff
@@ -0,0 +1,238 @@
...
+    @staticmethod
+    def parse_metrics_file(metrics_file):
+        data = defaultdict(dict)
+        with open(metrics_file) as f:
+            reader = csv.reader(f)
+            for line in reader:
+                if not line:
+                    continue
+
+                data[line[0]][line[2]] = line[3:]
```
lukbut: Why?

analysis_driver/pipelines/rapid.py
```diff
@@ -0,0 +1,238 @@
...
+
+    def _run(self):
+        data = []
+        for lane in sorted(self.dataset.rapid_samples_by_lane):
+            sample = self.dataset.rapid_samples_by_lane[lane]
+            user_sample_id = sample['User Sample Name']
+            output_dir = self.rapid_output_dir(lane)
+            map_metrics = self.parse_metrics_file(
+                util.find_file(output_dir, '%s.mapping_metrics.csv' % user_sample_id)
+            )['MAPPING/ALIGNING SUMMARY']
```
lukbut, line 103: I think this should be in a config file, in case it changes.

analysis_driver/pipelines/rapid.py
```diff
@@ -0,0 +1,238 @@
...
+
+            vc_metrics = self.parse_metrics_file(
+                util.find_file(output_dir, '%s.vc_metrics.csv' % user_sample_id)
+            )['VARIANT CALLER POSTFILTER']
```
lukbut, line 107: Same with this, I would be happier if this is a constant that is referred to from a config file.

analysis_driver/pipelines/rapid.py
```diff
@@ -0,0 +1,238 @@
...
+        )
+
+    def _run(self):
+        exit_status = 0
+
+        for lane in sorted(self.dataset.rapid_samples_by_lane):
+            sample = self.dataset.rapid_samples_by_lane[lane]
+            sample_id = sample['sample_id']
+
+            if not self.review(sample_id):
```
lukbut: Why?

analysis_driver/pipelines/rapid.py
```diff
@@ -0,0 +1,238 @@
...
+        os.makedirs(staging_dir, exist_ok=True)
+        transfer_data.create_output_links(
+            self.job_dir,
+            'rapid_analysis',
+            staging_dir,
+            user_sample_id=user_sample_id,
+            lane=lane
+        )
+
+        # TODO: implement generic MD5Sum stage
```
lukbut: Should this TODO be here, or is it better suited for a GitHub issue?

analysis_driver/dataset.py
```diff
@@ -344,6 +346,27 @@ def _find_pooling_step_for_artifact(art, expected_pooling_step_name=None, max_it
             )
         return art.input_artifact_list()
 
+    @property
+    def rapid_samples_by_lane(self):
+        if self._rapid_samples_by_lane is None:
+            self._rapid_samples_by_lane = {}
+
+            # TODO: move away from using the Lims API
```
lukbut: I believe this would be better placed in a GitHub issue than in a todo comment.

integration_tests/integration_test.py
```diff
@@ -71,21 +70,27 @@ def setUp(self):
...
         rest_communication.post_entry('projects', {'project_id': self.project_id, 'samples': [self.sample_id]})
-
-        self.dynamic_patches = []
         self._test_success = True
 
     def tearDown(self):
         super().tearDown()
         self._reset_logging()
-        for p in self.dynamic_patches:
-            p.stop()
```
lukbut, line 94: What did this do?

integration_tests/mocked_data.py
```diff
@@ -21,22 +16,22 @@ class MockedSample(NamedMock):
...
-                      samples=[MockedSample(real_name='10015AT0004', id='LP6002014-DTP_A04')]),
-            NamedMock(reagent_labels=['D705-D502 (ATTCAGAA-ATAGAGGC)'],
-                      samples=[MockedSample(real_name='10015AT0006', id='LP6002014-DTP_A05')]),
-            NamedMock(reagent_labels=['D706-D502 (GAATTCGT-ATAGAGGC)'],
-                      samples=[MockedSample(real_name='10015AT0007', id='LP6002014-DTP_A06')]),
-            NamedMock(reagent_labels=['D707-D502 (CTGAAGCT-ATAGAGGC)'],
-                      samples=[MockedSample(real_name='10015AT0008', id='LP6002014-DTP_A07')]),
-            NamedMock(reagent_labels=['D708-D502 (TAATGCGC-ATAGAGGC)'],
-                      samples=[MockedSample(real_name='10015AT0009', id='LP6002014-DTP_A08')])
+            Mock(reagent_labels=['D701-D502 (ATTACTCG-ATAGAGGC)'],
```
lukbut, line 19: What's the difference between NamedMock and Mock?

lukbut (commented):
This is a considerably large change and I believe more comments/documentation can help future reviewers/maintainers be more efficient in their work. 

Other than that well done, it looks like it took a lot of effort.

analysis_driver/dataset.py
```diff
@@ -344,6 +346,27 @@ def _find_pooling_step_for_artifact(art, expected_pooling_step_name=None, max_it
...
+            flowcell = set(self.lims_run.parent_processes()).pop().output_containers()[0]
+            for lane, artifact in flowcell.placements.items():
+                if len(artifact.reagent_labels) > 1:
+                    continue  # we don't want to run rapid processing on pools
+
+                assert len(artifact.samples) == 1
+                sample = artifact.samples[0]
+                if sample.udf.get('Rapid Analysis'):
```
mwhamgenomics: What data type will this UDF be? If it's a string, this will evaluate to `True` even if it's something like `'no'`

analysis_driver/pipelines/common.py
mwhamgenomics: The output file config no longer changes state, so this is not happening at all anymore. Instead of, e.g:
```python
output_file_config.set_pipeline_type('human sample processing')
vcf_stats = output_file_config['vcf_stats']
```
we now do:
```python
vcf_stats = output_file_config['human sample processing']['vcf_stats']
```

analysis_driver/pipelines/rapid.py
mwhamgenomics: These mirror the ulimits that get set automatically on the Dragen server in a login shell. There's an Illumina support ticket that discusses this

analysis_driver/pipelines/rapid.py
mwhamgenomics: This is very unlikely to change unless it's accompanied by other changes to the QC file that would require further code changes anyway.

analysis_driver/pipelines/rapid.py
mwhamgenomics: As above

integration_tests/integration_test.py
mwhamgenomics: `self.patches` is now a list not a tuple, i.e. is no longer immutable. This means we can now add extra patches to that, and the superclass calls `stop` on them automatically.

integration_tests/mocked_data.py
mwhamgenomics: NamedMock defines a property that works around `unittest.Mock` behaving weirdly if it's mocking an object that has an attribute called `name` - see tests/test_analysisdriver.py. These top level objects don't need to be NamedMocks

analysis_driver/pipelines/rapid.py
mwhamgenomics: See #274

analysis_driver/pipelines/rapid.py
mwhamgenomics: `self.review` returns True is a sample passed QC and False if it failed, and we don't want to output/deliver failing samples.

analysis_driver/pipelines/rapid.py
mwhamgenomics: I can't use DictReader because the QC file has no header. Will unpack the line into named variables.

analysis_driver/dataset.py
mwhamgenomics: @tcezard - what do we want to do here? This may require changes in the Rest API. The information we are currently getting from the Lims API consists of:
- lane number
- sample ID
- user sample ID
- project ID
- whether the sample is part of a pool (at the moment, determined by `len(reagent_labels) > 1`)
- number of samples in the lane (should always be 1)
- Rapid Analysis UDF

analysis_driver/pipelines/rapid.py
mwhamgenomics: This has now been done

analysis_driver/dataset.py
tcezard: I would split this in two:
 - The value we can get from the `lims/sample_info` endpoint
   - lane number
   - sample ID
   - project ID
   - user sample ID UDF, Rapid Analysis UDF (and other UDFs)
 - The rest can be obtained from the samplesheet (which still comes from the LIMS REST API)
   - whether the sample is part of a pool (at the moment, determined by len(reagent_labels) > 1)
   - number of samples in the lane (should always be 1)

mwhamgenomics: After discussion, there is a decision to be made - either of:
- take out all the Lims API queries (including advancing the Bioinformatics Processing step) and query the database instead (we'd need to add flowcell information to Reporting-App)
- keep the Lims API calls, and set up a dev instance of the Lims to be used in integration tests

The objective behind both of these is to stop patching the Lims in the integration tests. Removing Todo for now, pending further discussion.

analysis_driver/pipelines/rapid.py
lukbut: It might be valuable to trace that by linking the support ticket in, for future reference. What do you think @tcezard ?

analysis_driver/pipelines/rapid.py
lukbut: I see. A comment would be great here.

analysis_driver/pipelines/rapid.py
lukbut: Great, maybe we can discuss avoiding TODOs in the source code with @tcezard just to keep things tidy, or else including the issue number so there's better traceability? It's up for discussion.

analysis_driver/dataset.py
```diff
@@ -344,6 +346,31 @@ def _find_pooling_step_for_artifact(art, expected_pooling_step_name=None, max_it
...
+    @property
+    def rapid_samples_by_lane(self):
+        """
+        Search the sequencing run in the Lims for any samples with the UDF 'Rapid Analysis' set, and return their sample
+        ID, project ID and all UDFs.
+        """
+        if self._rapid_samples_by_lane is None:
+            self._rapid_samples_by_lane = {}
+
+            # TODO: move away from using the Lims API
```
tcezard: I agree with Luke's original comment that `TODO` by themselves do not allow us to track the tasks. 
I've open an issue to limit the use of the LIMS API, can you reference #394 in the `TODO`

analysis_driver/pipelines/rapid.py
tcezard: Can you add a reference to #274 in the TODO

tcezard (changes requested):

analysis_driver/pipelines/rapid.py
tcezard: I don't think a ticket would be very relevant here however a simple comment to state that these are the ulimit set on dragen server would be good.

analysis_driver/pipelines/rapid.py
```diff
@@ -0,0 +1,272 @@
...
+                   '--vc-sample-name {user_sample_id} --bcl-sample-sheet {sample_sheet} ' \
+                   '--enable-map-align-output true --enable-duplicate-marking true --dbsnp {dbsnp}; ' \
+
+            # somehow, writing the output data to scratch and then rsyncing to the main file system is faster than
+            # outputting directly
+            cmd += 'echo "Outputting data at $(date +%Y-%m-%d\ %H:%M:%S)"; ' \
+                   'rsync -rLD {tmp_dir}/ {out_dir}; rm -r {tmp_dir}; ' \
+                   'echo "Done at $(date +%Y-%m-%d\ %H:%M:%S)"'
+
+            cmds.append(
```
tcezard, line 40: All these command should be separated by `&&` instead of `;` so the script stops if any fails.
Alternatively you can set `set -e` at the start

tcezard (changes requested):
one small issue on the command.

analysis_driver/pipelines/rapid.py
mwhamgenomics: We want Dragen's exit status to be returned, but the cleanup afterwards always needs to happen otherwise the server will run out of scratch space. Adding `&&` for setup commands, but keeping `;` after Dragen and capturing/returning `$?`

tcezard (approved):

lukbut (approved):

analysis_driver/dataset.py
```diff
@@ -211,7 +211,11 @@ def _processing_instruction(self):
 
     def resolve_pipeline_and_toolset(self):
         instruction = self._processing_instruction()
-        toolset.configure(instruction['toolset_type'], instruction['toolset_version'])
+        toolset.configure(
+            instruction['toolset_type'],
+            instruction['toolset_version'],
+            os.path.join(cfg['jobs_dir'], self.name, 'program_versions.yaml')
+        )
```
tcezard, line 218: I have done that as well in my branch which is why I did not report it.

tcezard (approved):

lukbut (approved):

tcezard:
Validation ID: VB8

PR #389, opened 2019-03-27
IDT barcodes, more accurate tool version reporting, genome used

### LINKED ISSUE #373: Upload genome_used
In sample processing, we should upload the genome used when the sample starts processing.

### LINKED ISSUE #376: record version of tools used
At the moment we record the version of all the tools available is written to a yaml file at the end of the pipeline. As more tools gets added to the configuration file the number of tools in the yaml file will grow even if they are not used in the pipeline that ran.
We should record in the toolset the tool used in a particular run and only output these tool versions to the yaml files

### LINKED ISSUE #379: EGCG-Core#99 compatibility
EdinburghGenomics/EGCG-Core#99 will no longer include script parameter defaults if arguments are omitted. We should ensure all calls to `executor` are either local or set sensible defaults for CPU and memory, otherwise they will be set to Slurm defaults of 1 CPU and all available memory.

- common.py/bcbio_prepare_samples

### LINKED ISSUE #380: IDT library prep
The LIMS now supports IDT library prep adaptors as well as Illumina TruSeq. This breaks the regex in `RunDataset._run_elements_from_lims` because the reagent labels are different:

```
    run_element[ELEMENT_BARCODE] = match.group(3)                                                                                 
AttributeError: 'NoneType' object has no attribute 'group'
```

### LINKED ISSUE #381: interop metrics parser should not pass NaN
the interop summary can sometime output NaN in place of a value in its output.
The presence of NaN will make the JSON invalid so we need to avoid uploading these values.
The parser will be changed to replace NaN in the file with no entry being uploaded.

mwhamgenomics:
- Adds genome_used to MostRecentProc.start with SampleDataset - closes #373 
- Toolset now lazy-loads tools, making the report contain only the tools called in that pipeline run - closes #376 
- Makes all executor calls set `cpus` and `mem` explicitly - closes #379 
- Adds IDT barcode support - closes #380 
- Avoids uploading 'NaN' to Rest API in interop metrics - closes #381

### DIFF
--- a/analysis_driver/dataset.py
+++ b/analysis_driver/dataset.py
@@ -210,8 +210,7 @@ class Dataset(AppLogger):
 
     def resolve_pipeline_and_toolset(self):
         instruction = self._pipeline_instruction()
-        toolset.select_type(instruction['toolset_type'])
-        toolset.select_version(instruction['toolset_version'])
+        toolset.configure(instruction['toolset_type'], instruction['toolset_version'])
         self.pipeline = pipeline_register[instruction['name']]
 
 
--- a/analysis_driver/dataset.py
+++ b/analysis_driver/dataset.py
@@ -744,6 +744,9 @@ class MostRecentProc:
             if self.get('status') != DATASET_RESUME:
                 payload['start_date'] = now()
 
+            if self.dataset.type == 'sample':
+                payload['genome_used'] = self.dataset.genome_version
+
             self.update_entity(status=DATASET_PROCESSING, pid=os.getpid())
             rest_communication.patch_entry('analysis_driver_procs', payload, 'proc_id', self.proc_id)
 
--- a/analysis_driver/dataset.py
+++ b/analysis_driver/dataset.py
@@ -366,8 +366,23 @@ class RunDataset(Dataset):
                 if self.has_barcodes:
                     assert len(artifact.reagent_labels) == 1
                     reagent_label = artifact.reagent_labels[0]
-                    match = re.match('(\w{4})-(\w{4}) \(([ATCG]{8})-([ATCG]{8})\)', reagent_label)
-                    run_element[ELEMENT_BARCODE] = match.group(3)
+
+                    barcode = None
+                    for pattern in (
+                        # TruSeq label, e.g, A412-A208 (ATGCATGC-CTGACTGA)
+                        '(\w{4})-(\w{4}) \(([ATCG]{8})-([ATCG]{8})\)',
+                        # IDT label, e.g, 001A IDT-ILMN TruSeq DNA-RNA UD 96 Indexes Plate_UDI0001 (ATGCATGC-CTGACTGA)
+                        '(\w{4}) IDT-ILMN TruSeq DNA-RNA UD 96 Indexes (Plate_\w{7}) \(([ATGC]{8})-([ATGC]{8})\)'
+                    ):
+                        match = re.match(pattern, reagent_label)
+                        if match:
+                            barcode = match.group(3)
+
+                    if not barcode:
+                        raise AnalysisDriverError('Invalid reagent label found: %s' % reagent_label)
+
+                    run_element[ELEMENT_BARCODE] = barcode
+
                 run_elements.append(run_element)
         return run_elements
 
--- a/analysis_driver/quality_control/interop_metrics.py
+++ b/analysis_driver/quality_control/interop_metrics.py
@@ -27,22 +27,27 @@ class BadTileCycleDetector(AppLogger):
     def read_interop_metrics(self):
         run_metrics = RunMetrics()
         run_metrics.read(self.run_dir)
-        q_metric_set = run_metrics.q_metric_set()
-        metrics = q_metric_set.metrics()
+        metrics = run_metrics.q_metric_set().metrics()
 
         self.all_lanes = {}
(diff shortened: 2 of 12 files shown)


analysis_driver/dataset.py
```diff
@@ -729,6 +744,9 @@ def start(self):
             if self.get('status') != DATASET_RESUME:
                 payload['start_date'] = now()
 
+            if self.dataset.type == 'sample':
```
mwhamgenomics, line 772: I hate this so much.

analysis_driver/quality_control/interop_metrics.py
```diff
@@ -27,22 +27,27 @@ def __init__(self, dataset, window_size=60, tile_quality_threshold=None, cycle_q
...
-            cycle_dict[metrics[i].cycle()].append(metrics[i].qscore_hist())
+            tile_dict, cycle_dict = self.all_lanes[lane]
+            if tile not in tile_dict:
+                tile_dict[tile] = [None] * self.ncycles
+            if cycle not in cycle_dict:
+                cycle_dict[cycle] = []
+
+            qscore = m.qscore_hist()
+            tile_dict[tile][cycle-1] = qscore
+            cycle_dict[cycle].append(qscore)
```
mwhamgenomics, line 50: This refactor speeds up the reading of the InterOP metrics from ~8 seconds to ~3 seconds

analysis_driver/dataset.py
tcezard: It's either this or you need to duplicate the code in `SampleDataset`.
You could also extract a `_build_start_payload` from `start` where:
```python
def _build_start_payload(self):
    payload = {
        'pipeline_used': {
            'name': self.dataset.pipeline.name,
            'toolset_type': toolset.type,
            'toolset_version': toolset.version
        }
    }
    if self.get('status') != DATASET_RESUME:
        payload['start_date'] = now()
    return payload
```
Then in `SampleDataset`
```python
def _build_start_payload(self):
    payload = super()._build_start_payload(self)
    payload['genome_used'] = self.dataset.genome_version
    return payload
````

File: analysis_driver/tool_versioning.py [hunk 3]
```diff
@@ -122,6 +121,7 @@ def _get_stdout(cmd):
         return stdout.strip().decode('utf-8')
 
     def __getitem__(self, item):
+        self.add_tool(item)
```
tcezard: The issue with this implementation is that it also lazily check the version. 
If we have the wrong version it might crash 2 days into the run.
Not a major problem since we control the versions in use but still.

analysis_driver/dataset.py
mwhamgenomics: That is possible, but is more complex than the `if self.dataset.type == 'sample'`, which is ridiculous but simple - will try implementing it and see how I feel. I think this is related to EdinburghGenomics/Reporting-App#112, where we store history rather than state

analysis_driver/tool_versioning.py
mwhamgenomics: Also if we resume a sample, the tool versions file will be overwritten, potentially with missing tools from skipped stages. We could keep a register of all tools that a pipeline should use, but we'd then need to keep that up to date. Or we can remove the option of multiple pipelines sharing toolsets, and make it a 1 pipeline - 1 toolset relationship

analysis_driver/dataset.py
mwhamgenomics: Update: this is inside MostRecentProc, so we'd need to pass the payload down from SampleDataset, which is more trouble than it's worth. Current change is only 2 lines and reflects the craziness in the Reporting-App schema, so will keep it as it is.

analysis_driver/tool_versioning.py
mwhamgenomics: Update: toolset now loads, updates and rewrites the file if it already exists.

analysis_driver/reader/demultiplexing_parsers.py
```diff
@@ -394,22 +394,29 @@ def grab_sections(f):
                 sections[section_name].append(sp_line)
         return sections
 
+    def to_float(data):
+        n = float(data)
+        if math.isnan(n):
+            return None
+
```
lukbut, line 401: Extra blank space which can be removed.

analysis_driver/tool_versioning.py
```diff
@@ -26,10 +27,6 @@ def select_version(self, version):
         self.version = version
         self.tools = {}
         self.tool_versions = {}
-
-        for k in cfg['tools']:
-            self.add_tool(k)
```
lukbut, line 31: Where will the tools be added now that this loop has been removed?

mwhamgenomics: See line 132.

tcezard (changes requested):
I think you need to update `resolve_pipeline_and_toolset` in `dataset.py` to call `toolset. configure` instead of `toolset.select_type`

tcezard (changes requested):
Could you change one of the `reagent_label` in `integration_tests/mocked_data.py` to match an IDT barcode like in the unit test? 
You can leave the actual barcode the same as before so it won't have any side effect.

tcezard (approved):

tcezard:
validation ID: VB8

PR #391, opened 2019-04-08
Species/Genome is now loaded from the reporting app API

### LINKED ISSUE #367: Species/Genomes in the reporting app
- [x] Migrate from using a list of species/genomes from the config to using the ones stored in the reporting app. 
 - [x] Store the genome used during sample processing in analysis_driver_procs
 - [x] Enforce the the project whitelist, i.e. only process a sample if the project ID is in the whitelist, or the whitelist is missing.
 - [x] Remove support for "known_indels" in the genome reference data

Validation ID: VB8

lukbut:
Migrate from using a list of species/genomes from the config to using the ones stored in the reporting app.
Enforce the the project whitelist, i.e. only process a sample if the project ID is in the whitelist, or the whitelist is missing.

closes #367 
Validation ID: VB8


tests/test_pipelines/test_demultiplexing.py
```diff
@@ -24,8 +24,24 @@ def test_run(self):
...
+            "_created": "30_11_2018_15:13:43",
+            "species": "PhiX",
+            "genome_size": 5386,
+            "_id": "5c0153a716a5772f9e9cfdcc"
+        }
         patch_run_crawler = patch('analysis_driver.pipelines.demultiplexing.RunCrawler', autospec=True)
         patch_find = patch('egcg_core.util.find_all_fastq_pairs', side_effect=fake_fastq_pairs)
+        patch_get_document = patch('egcg_core.rest_communication.get_document',
+                                   return_value=fake_genome_response)
+        patch_get_document.start()
```
lukbut: This is a slightly different technique than the one you two currently use, which I found online. Let me know if you have any objections.

tests/test_pipelines/test_variant_calling.py
```diff
@@ -174,11 +205,10 @@ def test_run(self):
...
                                  '--read_filter BadCigar '
                                  '--read_filter NotPrimaryAlignment '
-                                 '-o tests/assets/jobs/test_dataset/gatk_var_calling/test_user_sample_id.intervals '
-                                 '-l INFO '
+                                 '-o tests/assets/jobs/test_dataset/gatk_var_calling/test_user_sample_id.intervals'
+                                 ' -l INFO '
                                  '-U LENIENT_VCF_PROCESSING '
-                                 '-I tests/assets/jobs/test_dataset/gatk_var_calling/test_user_sample_id_recal.bam '
-                                 '--known /path/to/known/indels',
+                                 '-I tests/assets/jobs/test_dataset/gatk_var_calling/test_user_sample_id_recal.bam',
```
lukbut, line 183: I was getting multiple assertion failures due to this whitespace. I do not understand why and I am not sure that moving it like I did is the best approach to solving this, but I cannot think of a better solution. Any tips welcome!

tests/test_pipelines/test_demultiplexing.py
tcezard: using `start()` and `stop()` only makes sense if the patch is started and stopped in different functions.
otherwise you should use `with`
There are no advantages to use `start()` and `stop()` here

lukbut: I was not sure how to patch it with "with" since this function is not explicitly called in test code. I will experiment with using "with" again to be sure of the difficulty I was facing.

tests/test_pipelines/test_variant_calling.py
```diff
@@ -174,11 +184,10 @@ def test_run(self):
                                  '-T RealignerTargetCreator '
                                  '--read_filter BadCigar '
                                  '--read_filter NotPrimaryAlignment '
-                                 '-o tests/assets/jobs/test_dataset/gatk_var_calling/test_user_sample_id.intervals '
-                                 '-l INFO '
+                                 '-o tests/assets/jobs/test_dataset/gatk_var_calling/test_user_sample_id.intervals'
+                                 ' -l INFO '
```
tcezard: Not that this is a real issue but I do not get any failure regardless of where the space is located here. Can you revert back?

analysis_driver/pipelines/variant_calling.py
```diff
@@ -84,15 +84,15 @@ def filter_snp_vcf(self):
 
     @property
     def dbsnp(self):
-        dbsnp = cfg.query('genomes', self.dataset.genome_version, 'dbsnp')
-        if not dbsnp and self.dataset.pipeline.name == 'variant_calling':
-            raise AnalysisDriverError('Could not find dbsnp file for %s' % self.dataset.name)
+        dbsnp = None
+        try:
+            dbsnp = cfg.get('genomes_dir', '') + \
+                    rest_communication.get_document('genomes', where={'assembly_name': self.dataset.genome_version})['data_files']['variation']
```
tcezard: I think I would prefer if all these `rest_communication` calls where all in `dataset.py`. You can cache the result of the query and even reconstruct the full path there. This will avoid further code duplication.

tcezard (changes requested):

tests/test_pipelines/test_variant_calling.py
lukbut: @tcezard I have reverted as instructed and the assertions started failing again. I have now only removed the trailing whitespace from the string defined on lines 190 and 215, as these strings are the last strings in the set of strings that together make up the command. Please let me know if you're happy with this.

analysis_driver/pipelines/variant_calling.py
lukbut: Good idea @tcezard. I'm going to create a function which does all this.

lukbut:
@tcezard This PR is ready for your review.

tcezard (approved):

PR #399, opened 2019-05-27
VB8 validation corrections

tcezard:
- Interop parsers: Ensure that not value are uploaded in case of NaN found in the file.
 -  Add tests for start of sample process due to coverage threshold reached
 -  Add tests for genome used being uploaded during processing
 - Add unit test of the project process using depagination to get the sample processed.


integration_tests/integration_test.py
```diff
@@ -336,15 +337,18 @@ def test_bcbio(self):
...
+            ad_proc['pipeline_used'],
             {'toolset_type': 'human_sample_processing', 'name': 'bcbio', 'toolset_version': 0},
             'pipeline used'
         )
+        self.expect_equal(ad_proc['genome_used'], 'hg38', 'genome used')
 
         self.expect_equal(
             rest_communication.get_document('analysis_driver_procs')['data_source'],
-            ['_'.join([self.run_id, str(i), self.barcode]) for i in range(1, 9)],
+            ['_'.join([self.run_id, str(i), self.barcode]) for i in range(1, 8)],
```
lukbut, line 351: Just out of interest, why did this decrease?

integration_tests/integration_test.py
```diff
@@ -562,6 +574,17 @@ def test_rapid_analysis(self):
 
     def test_gatk4_qc(self):
         self.setup_test('sample', 'test_gatk4_qc', 'gatk4_qc', 'Canis lupus familiaris', 'Not Variant Calling')
+
+        run_elements = []
+        # remove  Yield from the run elements and add coverage so it start because it passes the coverage threshold.
```
lukbut:
```suggestion
        # remove  Yield from the run elements and add coverage so it starts because it passes the coverage threshold.
```

lukbut (approved):

integration_tests/integration_test.py
tcezard: The first 7 lanes are marked as useable but lane 8 is not anymore (see line 67) and thus won't be used.

mwhamgenomics (approved):

PR #402, opened 2019-06-11
Pipeline classes

### LINKED ISSUE #390: Change AnalysisDriver pipelines into class
The pipeline are currently modules but they should be classes.
That would allow us to use inheritance to extend pipelines.

mwhamgenomics:
- Removes `pipeline_register`, Datasets now return pipeline objects
- Adds Pipeline superclass, makes pipelines classes

Closes #390

### DIFF
--- a/analysis_driver/pipelines/human_variant_calling_gatk4.py
+++ b/analysis_driver/pipelines/human_variant_calling_gatk4.py
@@ -1,18 +1,12 @@
 from egcg_core import executor
-from egcg_core.config import cfg
-
 from analysis_driver import quality_control as qc
-from analysis_driver.pipelines import common
-from analysis_driver.pipelines.common import MergeFastqs, SamtoolsStats
+from analysis_driver.segmentation import Parameter
+from analysis_driver.pipelines import Pipeline, common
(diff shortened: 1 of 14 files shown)


analysis_driver/pipelines/human_variant_calling_gatk4.py
```diff
@@ -139,74 +133,78 @@ def _run(self):
...
+
+        variant_file = gather_vcf.genotyped_vcf
+        steps_required = [gather_vcf]
+        if 'snpEff' in self.dataset.genome_dict:
+            # variant annotation
+            annotate_vcf = self.stage(VariantAnnotation, previous_stages=[gather_vcf])
+            variant_file = annotate_vcf.snps_effects_output_vcf
+            steps_required = [annotate_vcf]
+
+        # VQSR is disabled for now until we can make it perform better than hard filtering.
```
lukbut, line 177: I'd prefer it if this was traceable with a GitHub issue. Obviously given the circumstances...

lukbut (approved):

analysis_driver/pipelines/human_variant_calling_gatk4.py
tcezard: That one is actually on me. I'll put an issue up for it.

PR #406, opened 2019-06-25
Add support for extra space in the IDT barcode

### LINKED ISSUE #405: Fix IDT barcode parser
The IDT barcodes have a double space in their name which was not detected before.
We need to support it.

tcezard:
close #405

### DIFF
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -220,7 +220,7 @@ mocked_lane_artifact7 = NamedMock(real_name='art7', reagent_labels=['D706-D502 (
 mocked_lane_artifact8 = NamedMock(real_name='art8', reagent_labels=['D706-D502 (GAATTCGG-ATAGAGGC)'], samples=[MockedSample(real_name='sample8', udf={})])
 mocked_idt_artifact = NamedMock(
     real_name='idt_art',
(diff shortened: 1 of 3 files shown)


lukbut (approved):

tests/test_dataset.py
```diff
@@ -220,7 +220,7 @@ def test_pipeline_instruction(self):
 mocked_lane_artifact8 = NamedMock(real_name='art8', reagent_labels=['D706-D502 (GAATTCGG-ATAGAGGC)'], samples=[MockedSample(real_name='sample8', udf={})])
 mocked_idt_artifact = NamedMock(
     real_name='idt_art',
-    reagent_labels=['001A IDT-ILMN TruSeq DNA-RNA UD 96 Indexes Plate_UDI0001 (CCGCGGTT-AGCGCTAG)'],
+    reagent_labels=['001A IDT-ILMN TruSeq DNA-RNA UD 96 Indexes  Plate_UDI0001 (CCGCGGTT-AGCGCTAG)'],
```
mwhamgenomics, line 223: Unit tests still pass without this because of the `\s+` in the regex, so ideally we'd want a test for both one space and two spaces, but that would be a pain the way the tests are currently set up, so this should be fine

mwhamgenomics (approved):

PR #410, opened 2019-07-08
Use new toolset to update GATK version

### LINKED ISSUE #408: Support for GATK3.8
Temporary request to support GATK3.8 so we will add a new toolset for non-human variant calling and release in a separate tag.
This will be reverted in the subsequent version.

tcezard:
closes #408

### DIFF
--- a/etc/tool_versioning.yaml
+++ b/etc/tool_versioning.yaml
@@ -101,11 +101,51 @@ toolsets:
             tabix:
                 version: 1.3
                 version_cmd: grep_version
-        1:
+        -1:
+            bcbio:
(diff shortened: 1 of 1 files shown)


mwhamgenomics:
As discussed, this will work as is but will update GATK for all new projects, meaning we'd need to revalidate. A discussed fix would be to make the new toolset version `-1`, which is a bit janky and would not inherit all tools from version 0, because -1 < 0

mwhamgenomics (approved):
I don't particularly like it and there's definitely a better way of doing it, but we need this in order to process some stuff

PR #412, opened 2019-08-06
Asana update

mwhamgenomics:
Updates EGCG-Core - introduces Asana fixes implemented in EdinburghGenomics/EGCG-Core#109


requirements.txt
```diff
@@ -1,4 +1,4 @@
-EGCG-Core==0.11
+EGCG-Core==0.11.1
```
tcezard, line 1: That should be the newer version

```suggestion
EGCG-Core==0.11.2
```

tcezard (changes requested):

PR #416, opened 2019-09-06
Fix fastqc command

### LINKED ISSUE #415: fastqc fails silently
By default fastqc uses a temp directory that is located in /tmp. In some cases this directory does not have enough space to host the html output which makes fastqc fails.
Despite the failure, fastqc still reports a 0 exist status.
We should:
 - Change the temp directory in the fastqc command line to a directory in the job folder.
 - Ensure that the html and zip output exists at the end of the command and return non 0 exist if they don't.

tcezard:
Fix the fastqc command so it uses a local directory for its temp directory
closes #415

### DIFF
--- a/analysis_driver/util/bash_commands.py
+++ b/analysis_driver/util/bash_commands.py
@@ -27,8 +27,10 @@ def bcl2fastq(input_dir, fastq_path, sample_sheet=None, mask=None):
     return cmd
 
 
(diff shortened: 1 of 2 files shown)


lukbut (approved):

tcezard:
This PR passed the integration tests.

mwhamgenomics (approved):

PR #419, opened 2019-09-18
394 limit lims rest api

### LINKED ISSUE #394: Limit the use of LIMS REST API
The usage of LIMS REST API is difficult to test integration tests because we do not have a test LIMS. Our integration tests mock all interaction with the LIMS which means any change is hazardous.
We should limit the use of calls to the LIMS REST API and when it is necessary we should refactor the code so it is only used in one place.

tcezard:
Use our REST API lims endpoint instead of the LIMS REST API in dataset.
Specify data to be loaded in the LIMS database for integration tests.
It involves lots of changes in the integration test but not that many changes in the actual code
close #394


analysis_driver/dataset.py
```diff
@@ -723,15 +730,12 @@ def species(self):
     @property
     def genome_version(self):
         if self._genome_version is None:
-            g = clarity.get_sample(self.samples_processed[0]['sample_id']).udf.get('Genome Version')
-            if not g:
-                g = rest_communication.get_document('species', where={'name': self.species})['default_version']
-            self._genome_version = g
+            self._genome_version = self.sample_dataset(self.samples_processed[0]['sample_id']).genome_version
```
lukbut, line 747: Can you please point me to where this information is now downloaded, since the "get_document" call has been removed?

analysis_driver/dataset.py
```diff
@@ -351,63 +353,55 @@ def rapid_samples_by_lane(self):
...
+            for sample_data in lane['samples']:
                 run_element = {
-                    ELEMENT_PROJECT_ID: sample.project.name,
-                    ELEMENT_SAMPLE_INTERNAL_ID: sample.name,
-                    ELEMENT_LIBRARY_INTERNAL_ID: sample.id,  # This is not the library id but it is unique
-                    ELEMENT_LANE: lane.split(':')[0],
+                    ELEMENT_PROJECT_ID: sample_data['project_id'],
+                    ELEMENT_SAMPLE_INTERNAL_ID: sample_data['sample_id'],
+                    ELEMENT_LIBRARY_INTERNAL_ID: sample_data['artifact_id'],  # This is not the library id but it is unique
+                    ELEMENT_LANE: str(lane['lane']),
```
mwhamgenomics, line 384: `lane` is an integer in Reporting-App's schema.yaml

analysis_driver/dataset.py
mwhamgenomics: The logic here is identical to SampleDataset.genome_version at line 500, so ProjectDataset.genome_version just goes through that now

tests/test_dataset.py
```diff
@@ -295,60 +295,37 @@ def setup_dataset(self):
...
+        with patch('egcg_core.rest_communication.get_document', return_value={}):
+            assert self.dataset.rapid_samples_by_lane == {}
+        responses = [{},
+                     {'sample_id': 'sample2', 'Rapid Analysis': 'Yes', 'project_id': '10015AT'},
+                     {}, {}, {}, {}, {}, {}]
+        with patch('egcg_core.rest_communication.get_document', side_effect=responses):
+            self.dataset._rapid_samples_by_lane = None
+            assert self.dataset.rapid_samples_by_lane == {'2': {'sample_id': 'sample2', 'Rapid Analysis': 'Yes', 'project_id': '10015AT'}}
 
-        self.dataset._rapid_samples_by_lane = None
```
mwhamgenomics, line 292: I think this now conflicts since I merged #420

mwhamgenomics (changes requested):

analysis_driver/dataset.py
mwhamgenomics: The logic here is identical to SampleDataset.genome_version at line 500, so ProjectDataset.genome_version just goes through that now

analysis_driver/dataset.py
tcezard: True but it seems like it has always been a string at this point because it came from the artifact position in the flowcell `'1:1'` which got parsed.
If I don't cast it here it fail to find the the lane at 
https://github.com/EdinburghGenomics/Analysis-Driver/blob/199cb668fccc63c95bc7e148722076d82948d527/analysis_driver/report_generation/run_crawler.py#L271
I could have cast it there but I prefer to keep backward compatibility to avoid having other issues elsewhere.
The upload to the reporting-app force the cast to `int`

tests/test_dataset.py
tcezard: I've merged the release branch and resolved conflicts. This one did not show up.

mwhamgenomics (approved):

lukbut (approved):

PR #422, opened 2019-10-07
PDP pool fix

mwhamgenomics:
Lets `RunDataset._find_pooling_step_for_artifact` check for multiple step names, adds Create PDP Pool EG 1.0 ST to list of checked names.


tcezard (approved):

tcezard (commented):
I just realised that this part of the code is being rewritten in https://github.com/EdinburghGenomics/Analysis-Driver/pull/419 
so this code won't be used unless we release a intermediate version

lukbut (approved):

PR #423, opened 2019-11-04
Split demultiplexing per lane

### LINKED ISSUE #411: Split bcl2fastq job
bcl2fastq can be run on single lane which mean the job can be made considerably faster by splitting it over multiple process.
This should only involve adding an option to the bcl2fastq command to specify the lane and iterate over each lane.
We need to ensure the result do not change the fastq files.

### LINKED ISSUE #417: demultiplexing of single sample in pooling run
When a pooling run is processed, if a single sample is on a lane it will be demultiplexed with the provided barcode and the undetermined fastq will be lost.

We need to allow this case to be processed as non pooling. 
This is linked to #411 because the lanes need to be processed separately

tcezard:
This PR split the demultiplexing per lane and allow to demultiplex pooling and non pooling lane in the same run.

Lots of files are just a copy of the existing run for the `test_split_run_dir` asset

closes #411 
closes #417


analysis_driver/dataset.py
```diff
@@ -401,8 +407,17 @@ def _run_elements_from_lims_endpoint(self):
...
+        """
+        Returns False if this lane has only one barcode (regardless of what is in the RunInfo.xml).
+        Otherwise delegate to the RunInfo object.
+        :param lane_number: the lane number to query.
+        :return: True if this lane needs to be use the barcode False otherwise.
+        """
+        lane_data = [l for l in self.run_status['lanes'] if str(l['lane']) == str(lane_number)][0]
+        # If there is only one sample in this lane then we won't demultiplex and should use a barcode length of 0
+        if len(lane_data['samples']) == 1:
+            return False
```
mwhamgenomics, line 420: A unit test for this behaviour would be good

analysis_driver/report_generation/crawler.py
```diff
@@ -7,18 +7,18 @@
...
-        if 'Required Yield (Gb)' in lims_sample.udf:
-            sample_info[c.ELEMENT_SAMPLE_REQUIRED_YIELD] = lims_sample.udf.get('Required Yield (Gb)') * 1000000000
-        if 'Coverage (X)' in lims_sample.udf:
-            sample_info[c.ELEMENT_SAMPLE_REQUIRED_COVERAGE] = lims_sample.udf.get('Coverage (X)')
+        if 'Yield for Quoted Coverage (Gb)' in rest_data:
+            sample_info[c.ELEMENT_SAMPLE_REQUIRED_YIELD_Q30] = int(rest_data.get('Yield for Quoted Coverage (Gb)')) * 1000000000
+        if 'Required Yield (Gb)' in rest_data:
+            sample_info[c.ELEMENT_SAMPLE_REQUIRED_YIELD] = int(rest_data.get('Required Yield (Gb)')) * 1000000000
+        if 'Coverage (X)' in rest_data:
+            sample_info[c.ELEMENT_SAMPLE_REQUIRED_COVERAGE] = rest_data.get('Coverage (X)')
```
mwhamgenomics: The Lims endpoint returns a string here - don't we want to cast this as well?

analysis_driver/report_generation/run_crawler.py
```diff
@@ -103,7 +103,7 @@ def _populate_barcode_info_from_dataset(self, dataset):
         for project_id in self.projects:
             self.projects[project_id][ELEMENT_SAMPLES] = list(self.projects[project_id][ELEMENT_SAMPLES])
 
-        if dataset.has_barcodes:
+        if dataset.has_barcode_in_lane(run_element[ELEMENT_LANE]):
```
mwhamgenomics: Here we will process only the last run element iterated over in `dataset.run_elements` - is this really what we want?

analysis_driver/report_generation/run_crawler.py
```diff
@@ -217,9 +217,17 @@ def _populate_barcode_info_from_json_file(self, run_dir):
...
+            json_files = util.find_files(run_dir, 'Stats', 'lane_*_Stats.json')
+
+        if not json_files:
+            # Either the full or split file are expected.
+            # The process should stop if they are not found.
+            raise FileNotFoundError()
+
+        for json_file in json_files:
+            with open(json_file, 'r') as open_file:
+                json_data = json.load(open_file)
```
mwhamgenomics, line 230: Here we are iterating over different JSON files split by lane, and then running code that populates info based on run element ID. Was wondering if we could end up overwriting data in `self.barcodes_info`, but since run element ID depends on the lane number, I think this should be fine?

mwhamgenomics (changes requested):

analysis_driver/util/helper_functions.py
```diff
@@ -50,3 +54,56 @@ def split_in_chunks(total_length, chunksize, zero_based=False, end_inclusive=Tru
         chunks.append((last, min(new, total_length + end_padding)))
         last = new + chunk_padding
     return chunks
+
+
+def merge_lane_directories(fastq_dir, run_elements):
+    """
+    At the end of a per lane bcl2fastq run the data in the lane directories must be merged back to the main fastq dir
+    This function does that by moving all the fastq files to the expected location
+    and the stats file are move to a central Stats folder. They are rename to avoid file collision.
```
lukbut:
```suggestion
    and the stats file are move to a central Stats folder. They are renamed to avoid file collision.
```

analysis_driver/report_generation/run_crawler.py
tcezard: Very good catch. That would have caused mixed flowcell to not have the unassigned run element if the last lane was not a pool

analysis_driver/report_generation/run_crawler.py
tcezard: If the json files had metrics for overlapping lanes (lane 1 and 2 in first one then lane 2 in the second one) then that could be a problem but it should not happen.
In any case the metrics could be overwritten but not the `self.barcodes_info` because this is only defined in `_populate_barcode_info_from_dataset`

lukbut (approved):

mwhamgenomics (approved):

PR #425, opened 2019-11-15
Fix bcl2fastq per lane

tcezard:
bcl2fastq does not support having barcodes in  the samplesheet when demultiplexing non pooling lanes. 
This is true regardless of if the lane being demultiplexed is non-pooling.
This means that the Samplesheet needs to be generated per lane rather than one per run.
This PR changes this behaviour and also create a test in the demultiplexing integration test with a mixture of pooling and non pooling lane.

### DIFF
--- a/analysis_driver/util/bash_commands.py
+++ b/analysis_driver/util/bash_commands.py
@@ -8,11 +8,11 @@ from analysis_driver.exceptions import AnalysisDriverError
 app_logger = log_cfg.get_logger(__name__)
 
 
-def bcl2fastq_per_lane(input_dir, fastq_dir, sample_sheet, masks, lanes):
+def bcl2fastq_per_lane(input_dir, fastq_dir, sample_sheet_func, masks, lanes):
     cmds = []
     for mask, lane in zip(masks, lanes):
         lane_fastq_dir = os.path.join(fastq_dir, 'lane_' + str(lane))
-        cmds.append(bcl2fastq(input_dir, lane_fastq_dir, sample_sheet, mask, lane))
+        cmds.append(bcl2fastq(input_dir, lane_fastq_dir, sample_sheet_func(lane), mask, lane))
     return cmds
 
 
--- a/analysis_driver/util/bash_commands.py
+++ b/analysis_driver/util/bash_commands.py
@@ -9,6 +9,7 @@ app_logger = log_cfg.get_logger(__name__)
 
 
 def bcl2fastq_per_lane(input_dir, fastq_dir, sample_sheet_func, masks, lanes):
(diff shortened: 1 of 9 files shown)


integration_tests/data_for_clarity_lims.yaml
```diff
@@ -53,7 +53,7 @@ artifacts:
     - { name: l4, samples: [10015AT0001, 10015AT0002, 10015AT0003, 10015AT0004, 10015AT0006, 10015AT0007, 10015AT0008, 10015AT0009], container_name: FLOWCELL1, xpos: 0, ypos: 3 }
     - { name: l5, samples: [10015AT0001, 10015AT0002, 10015AT0003, 10015AT0004, 10015AT0006, 10015AT0007, 10015AT0008, 10015AT0009], container_name: FLOWCELL1, xpos: 0, ypos: 4 }
     - { name: l6, samples: [10015AT0001, 10015AT0002, 10015AT0003, 10015AT0004, 10015AT0006, 10015AT0007, 10015AT0008, 10015AT0009], container_name: FLOWCELL1, xpos: 0, ypos: 5 }
-    - { name: l7, samples: [10015AT0001, 10015AT0002, 10015AT0003, 10015AT0004, 10015AT0006, 10015AT0007, 10015AT0008, 10015AT0009], container_name: FLOWCELL1, xpos: 0, ypos: 6 }
+    - { name: l7, samples: [10015AT0001], container_name: FLOWCELL1, xpos: 0, ypos: 6 }
     - { name: l8, samples: [10015AT0001, 10015AT0002, 10015AT0003, 10015AT0004, 10015AT0006, 10015AT0007, 10015AT0008, 10015AT0009], container_name: FLOWCELL1, xpos: 0, ypos: 7 }
```
tcezard, line 57: This lines means that lane 7 in the run will only contain `10015AT0001` which mean it is non-polling

analysis_driver/util/bash_commands.py
```diff
@@ -8,11 +8,12 @@
 app_logger = log_cfg.get_logger(__name__)
 
 
-def bcl2fastq_per_lane(input_dir, fastq_dir, sample_sheet, masks, lanes):
+def bcl2fastq_per_lane(input_dir, fastq_dir, sample_sheet_func, masks, lanes):
+    # Note sample_sheet_func is the function to get the samplesheet not the samplesheet file itself
```
tcezard, line 12: This is a bit of a hack. I didn't want to change where the iteration over 8 lanes happened so had to receive a function that generate the samplesheet instead of the samplesheet itself.

mwhamgenomics (approved):

lukbut (approved):

